PR #81, opened 2026-03-05
fix: credential masking panic and silent token write errors

0xMiguel:
## Summary

- **`auth_commands.rs`** — `gws auth export` credential masking had two bugs:
  1. **Panic on short strings**: `&s[..4]` causes an index-out-of-bounds panic when `client_secret` or `refresh_token` is shorter than 4 characters.
  2. **Masking doesn't actually mask**: `&s[s.len().min(4)..]` evaluates to `&s[4..]` for any string longer than 4 chars, which means the "masked" output contains **every character** of the secret (just with `...` inserted after position 4). The intent was `s.len().saturating_sub(4)` to show only the last 4 characters. Replaced both call sites with a `mask_secret()` helper that safely shows the first 4 and last 4 chars, or `"***"` for short strings.

- **`token_storage.rs`** — `save_to_disk` silently discarded the `io::Result` from `atomic_write_async` via `let _`, so the function returned `Ok(())` even when the write failed (e.g. disk full, permission denied). Token persistence failures now properly propagate via `?`.

## Test plan

- [x] `cargo test` passes
- [x] `gws auth export` with a short `client_secret` no longer panics
- [x] `gws auth export` output properly masks secrets (e.g. `ABCD...WXYZ` instead of `ABCD...EFGHIJKLMNOPQRSTUVWXYZ`)
- [x] Token cache write failures are now surfaced to the user

### DIFF
--- a/.changeset/fix-credential-masking.md
+++ b/.changeset/fix-credential-masking.md
@@ -1,0 +1,9 @@
+---
+"gws": patch
+---
+
+fix: credential masking panic and silent token write errors
+
+Fixed `gws auth export` masking which panicked on short strings and showed
+the entire secret instead of masking it. Also fixed silent token cache write
+failures in `save_to_disk` that returned `Ok(())` even when the write failed.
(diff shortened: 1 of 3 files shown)


jpoehnelt:
Start with the CLA and then I'll take a look at the rest.

PR #109, opened 2026-03-05
fix: replace unwrap() calls with proper error handling in MCP server

jeftekhari:
## Description                                                                                     
                                                                                                     
  Replace four `serde_json::to_string().unwrap()` calls in the MCP server (`src/mcp_server.rs`) that 
  could panic the server process if serialization ever fails. Instead:

  - **Response serialization** (2 instances): Log the error to stderr and `continue` to the next
  request gracefully
  - **Params/body serialization** (2 instances): Propagate errors via `?` using
  `Option::transpose()`, returning a proper `GwsError::Validation` to the caller
  - **Auth fallback**: Log a warning when authentication fails and the server silently falls back to
  unauthenticated mode, so users/agents are aware of degraded state

  ## Checklist:

These serialization paths are effectively infallible for well-formed serde_json::Value inputs,
  making them difficult to unit test in isolation. The change is defensive — replacing panics with
  graceful error handling for robustness.

### DIFF
--- a/src/mcp_server.rs
+++ b/src/mcp_server.rs
@@ -388,13 +388,13 @@ async fn handle_tools_call(params: &Value, config: &ServerConfig) -> Result<Valu
 
     let params_json_val = arguments.get("params");
     let params_str = params_json_val
-        .map(|v| serde_json::to_string(v))
+        .map(serde_json::to_string)
         .transpose()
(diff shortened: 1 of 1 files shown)


jpoehnelt:
Fix the clippy warning and 👍

jeftekhari:
> Fix the clippy warning and 👍

should be fixed!

PR #85, opened 2026-03-05
feat: multi-account support

jpoehnelt:
## Summary

Add multi-account support for the gws CLI with per-account credential storage. #78 

### New features
- `--account EMAIL` global flag available on every command
- `GOOGLE_WORKSPACE_CLI_ACCOUNT` environment variable as fallback
- `gws auth login --account EMAIL` — login and register a specific account
- `gws auth list` — list all registered accounts
- `gws auth default EMAIL` — set the default account
- `gws auth logout --account EMAIL` — remove a specific account
- `login_hint` in OAuth URL for automatic account pre-selection in browser
- Email validation via Google userinfo endpoint after OAuth flow

### Architecture
- Per-account encrypted credentials: `credentials.<b64-email>.enc`
- New `accounts.json` registry with default tracking
- New `src/accounts.rs` module (email normalization, base64 encoding, registry CRUD)
- Updated `src/credential_store.rs` with per-account path/save/load functions
- Updated `src/auth.rs` with account resolution logic

### Breaking change

> [!CAUTION]
> Existing users must run `gws auth login` again after upgrading. The credential storage format changed from a single `credentials.enc` to per-account files.

### Testing
- 380 tests pass (12 new unit tests)
- Tests cover: extract_global_account, filter_args_for_subcommand, encrypted_credentials_path_for, email normalization, base64 round-trip, registry CRUD

### Documentation
- README: multi-account section, updated precedence table
- AGENTS.md: new source files, env var
- Shared SKILL.md: --account flag, multi-account auth commands
- Changeset included

PR #117, opened 2026-03-05
fix: clean up per-account token caches on logout

adhishthite:
## Summary

Addresses review findings from #85 — logout now removes per-account token cache files alongside credential files.

### Problem

Both `gws auth logout --account EMAIL` and `gws auth logout` (full) removed encrypted credential files but left per-account token cache files on disk:
- `token_cache.<b64-email>.json`
- `sa_token_cache.<b64-email>.json`

This means cached OAuth tokens (potentially with valid access/refresh tokens) persist after explicit logout — a security concern and a correctness issue.

### Fix

- **Per-account logout** now removes `credentials.<b64>.enc`, `token_cache.<b64>.json`, and `sa_token_cache.<b64>.json` for the target account
- **Full logout** now iterates all registered accounts and removes their token cache files in addition to credential files

### Changes

- `src/auth_commands.rs`: Updated `handle_logout` to clean up per-account token caches in both single-account and full logout paths

### Test plan

- [x] `cargo check` — compiles clean
- [x] `cargo test` — all 390 tests pass
- Verified correct file path construction matches the pattern used in `auth.rs:76-83` for token cache creation

### DIFF
--- a/src/auth_commands.rs
+++ b/src/auth_commands.rs
@@ -1050,15 +1050,23 @@ fn handle_logout(args: &[String]) -> Result<(), GwsError> {
     }
 
     if let Some(ref email) = account_email {
-        // Per-account logout: remove only that account's credentials
+        // Per-account logout: remove only that account's credentials and token caches
         let enc_path = credential_store::encrypted_credentials_path_for(email);
+        let normalised = crate::accounts::normalize_email(email);
+        let b64 = crate::accounts::email_to_b64(&normalised);
+        let cfg = config_dir();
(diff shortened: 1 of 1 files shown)


adhishthite:
@jpoehnelt This is a small fix for your multi-account branch — logout was leaving per-account token cache files on disk. Happy to adjust if you'd prefer a different approach.

jpoehnelt:
Did you have any other issues with the multi-account branch?

adhishthite:
For now, none. :)

adhishthite:
@jpoehnelt do you need me to resolve the conflicts here?

PR #124, opened 2026-03-05
Fix/warn credential permission failures

jeftekhari:
## Description

Replaced silent `let _ =` on `set_permissions` calls in `save_encrypted()` with `eprintln!` warnings so users are aware if their credential files end up with insecure permissions. Also log keyring access failures instead of silently falling through to file storage.

This follows the existing pattern already used in `get_or_create_key()` (lines 92-98) where permission failures are properly warned about.

**Changes:**
- `save_encrypted`: Warn if directory permissions (0o700) fail to set
- `save_encrypted`: Warn if credentials file permissions (0o600) fail to set
- `get_or_create_key`: Log keyring errors instead of silently ignoring them

## Checklist:

### DIFF
--- a/src/credential_store.rs
+++ b/src/credential_store.rs
@@ -119,7 +119,9 @@ fn get_or_create_key() -> anyhow::Result<[u8; 32]> {
 
                 return Ok(cache_key(key));
             }
-            Err(_) => {} // Fallthrough to file storage
+            Err(e) => {
+                eprintln!("Warning: keyring access failed, falling back to file storage: {e}");
+            }
         }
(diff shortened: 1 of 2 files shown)


jpoehnelt:
Please fix the conflicts.

src/credential_store.rs
jeftekhari, line 1: > Please fix the conflicts.

it looks like someone has already added a fix for this. this is the only remaining bit from my code and it seems exceptionally minor.

jpoehnelt:
tested locally

PR #125, opened 2026-03-05
feat: add Application Default Credentials (ADC) support

### LINKED ISSUE #103: Support ADC
> This eliminates the need to create and share custom OAuth clients.

Note that most Workspace APIs will not be able to piggyback on the `gcloud` OAuth. However, ADC can be setup with a custom OAuth client and this library should support this pattern.

zerone0x:
## Description

Closes #103

Adds Application Default Credentials (ADC) as a 4th credential source in `get_token()`.

### New credential lookup order

| Priority | Source |
|---|---|
| 0 | `GOOGLE_WORKSPACE_CLI_TOKEN` env var (raw token) |
| 1 | `GOOGLE_WORKSPACE_CLI_CREDENTIALS_FILE` env var |
| 2 | Encrypted credentials (`~/.config/gws/credentials.enc`) |
| 3 | Plaintext credentials (`~/.config/gws/credentials.json`) |
| **4** | **ADC — `GOOGLE_APPLICATION_CREDENTIALS` env var, then `~/.config/gcloud/application_default_credentials.json`** |

### Supported ADC flows

This unlocks two common setups:

1. **User OAuth via gcloud ADC** (recommended for personal accounts):
   ```bash
   gcloud auth application-default login --client-id-file=client_secret.json
   # No need to run `gws auth login` separately
   ```

2. **Service account via GOOGLE_APPLICATION_CREDENTIALS**:
   ```bash
   export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json
   gws drive files list
   ```

Both `authorized_user` and `service_account` ADC formats are detected automatically via the `type` field.

As noted in the issue, most Workspace APIs require scopes granted to a specific user OAuth client, so service account ADC coverage is limited to APIs that support domain-wide delegation. User OAuth ADC with a custom client (via `--client-id-file`) works for all standard Workspace APIs.

**Dry Run Output:**
```json
// Not applicable — credential loading logic change, no HTTP request generated
```

## Checklist:

- [ ] I have run `cargo fmt --all` to format the code perfectly. *(no Rust toolchain on build host — CI will verify)*
- [ ] I have run `cargo clippy -- -D warnings` and resolved all warnings. *(CI will verify)*
- [x] I have added tests that prove my fix is effective or that my feature works. (`test_load_credentials_adc_env_var_authorized_user`, `test_load_credentials_adc_env_var_missing_file`)
- [x] I have provided a Changeset file to document my changes.

---
🤖 Generated with Claude Code

### DIFF
--- a/.changeset/adc-support.md
+++ b/.changeset/adc-support.md
@@ -7,11 +7,12 @@ Add Application Default Credentials (ADC) support.
 `gws` now discovers ADC as a fourth credential source, after the encrypted
 and plaintext credential files. The lookup order is:
 
-1. `GOOGLE_WORKSPACE_CLI_CREDENTIALS_FILE` env var
-2. Encrypted credentials (`~/.config/gws/credentials.enc`)
-3. Plaintext credentials (`~/.config/gws/credentials.json`)
-4. **ADC** — `GOOGLE_APPLICATION_CREDENTIALS` env var, then
-   `~/.config/gcloud/application_default_credentials.json`
+1. `GOOGLE_WORKSPACE_CLI_TOKEN` env var (raw access token, highest priority)
+2. `GOOGLE_WORKSPACE_CLI_CREDENTIALS_FILE` env var
+3. Encrypted credentials (`~/.config/gws/credentials.enc`)
+4. Plaintext credentials (`~/.config/gws/credentials.json`)
+5. **ADC** — `GOOGLE_APPLICATION_CREDENTIALS` env var (hard error if file missing), then
+   `~/.config/gcloud/application_default_credentials.json` (silent if absent)
 
 This means `gcloud auth application-default login --client-id-file=client_secret.json`
 is now a fully supported auth flow — no need to run `gws auth login` separately.
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -26,8 +26,15 @@ use crate::credential_store;
 
 /// Returns the well-known Application Default Credentials path:
 /// `~/.config/gcloud/application_default_credentials.json`.
+///
+/// Note: `dirs::config_dir()` returns `~/Library/Application Support` on macOS, which is
+/// wrong for gcloud. The Google Cloud SDK always uses `~/.config/gcloud` regardless of OS.
 fn adc_well_known_path() -> Option<PathBuf> {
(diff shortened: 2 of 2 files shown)


zerone0x:
good catches, updated:

- extracted the JSON parsing into a helper (`parse_credential_file`) so the env file and ADC paths share the same logic, also switched to `from_value` to skip the second string parse
- fixed the macOS path — `dirs::config_dir()` gives `~/Library/Application Support` on mac which is wrong, now using `home_dir().join(".config/gcloud/...")`
- `GOOGLE_APPLICATION_CREDENTIALS` pointing to a missing file now hard-errors instead of silently falling through
- added the missing service_account test for ADC
- removed the `unsafe` blocks, they were misleading
- updated the changeset to include `GOOGLE_WORKSPACE_CLI_TOKEN` at the top

jpoehnelt:
Re: the Gemini review concern about `--account` silently falling through to ADC — after tracing the code, this is a false positive. `resolve_account()` already hard-errors when an explicit `--account` is passed and the account isn't in the registry (both the `(Some, Some)` and `(Some, None)` arms bail). ADC is only reachable via the `(None, None)` path — no `--account` flag, no registry, no legacy creds — which is exactly the "fresh install with only gcloud configured" scenario where ADC should kick in.

jpoehnelt:
Re: the Gemini review concern about `--account` silently falling through to ADC — this is a false positive. `resolve_account()` already hard-errors when an explicit `--account` is passed and the account isn't in the registry (both the `(Some, Some)` and `(Some, None)` arms bail). ADC is only reachable via the `(None, None)` path — no `--account` flag, no registry, no legacy creds — which is the fresh-install-with-gcloud scenario where ADC should kick in.

jpoehnelt:
Re: the Gemini review concern about `--account` silently falling through to ADC — this is a false positive. `resolve_account()` already hard-errors when an explicit `--account` is passed and the account isn't in the registry (both the `(Some, Some)` and `(Some, None)` arms bail). ADC is only reachable via the `(None, None)` path — no `--account` flag, no registry, no legacy creds — which is the fresh-install-with-gcloud scenario where ADC should kick in.

Issue #119, opened 2026-03-05
`gws auth login` unusable with personal @gmail.com accounts — invalid_scope + ignores custom client_secret.json


shaun-agent:
## Summary

`gws auth login` (v0.3.4) is completely unusable with personal `@gmail.com` accounts. Two separate issues compound to make authentication impossible:

1. The "Recommended" scope preset includes admin-only scopes that Google rejects with `400 invalid_scope` for non-Workspace accounts
2. The CLI ignores a custom OAuth client placed at `~/.config/gws/client_secret.json` and always uses its hardcoded built-in client ID

## Environment

- **gws version**: 0.3.4 (npm install)
- **OS**: macOS (Apple Silicon)
- **Account type**: Personal `@gmail.com` (not Google Workspace)
- **GCP project**: Has all relevant APIs enabled (Calendar, Drive, Gmail, Docs, Sheets, Forms, Chat, etc.)

### Issue 1: invalid_scope with Recommended preset

1. Run `gws auth login`
2. Select "Recommended" scope preset (or any preset that includes admin scopes)
3. Browser opens Google OAuth consent screen
4. Google returns `400 invalid_scope`

The scope list sent to Google includes these admin-only scopes that are invalid for personal accounts:

```
https://www.googleapis.com/auth/apps.alerts
https://www.googleapis.com/auth/apps.groups.settings
https://www.googleapis.com/auth/apps.licensing
https://www.googleapis.com/auth/apps.order
https://www.googleapis.com/auth/cloud-identity.devices
https://www.googleapis.com/auth/cloud-identity.groups
https://www.googleapis.com/auth/cloud-identity.inboundsso
https://www.googleapis.com/auth/cloud-identity.policies
https://www.googleapis.com/auth/ediscovery
https://www.googleapis.com/auth/directory.readonly
https://www.googleapis.com/auth/groups
https://www.googleapis.com/auth/chat.admin.memberships
https://www.googleapis.com/auth/chat.admin.spaces
```

These scopes require Google Workspace Admin SDK / domain-wide delegation and will always fail for `@gmail.com` accounts.

### Issue 2: Custom client_secret.json ignored

1. Create a Desktop OAuth client in Google Cloud Console
2. Download the JSON and save to `~/.config/gws/client_secret.json`
3. Run `gws auth login`
4. The OAuth URL still uses the built-in client ID (`725852102922-v9qj7...`) instead of the custom one

[...]

PR #134, opened 2026-03-05
fix: use ~/.config/gws on all platforms for consistent config path

haunchen:
## Description

Partially addresses #119 (Bug 2: custom `client_secret.json` ignored)

### Problem

`config_dir()` used `dirs::config_dir()` which resolves to different paths per OS:

| Platform | `dirs::config_dir()/gws/` | README says |
|----------|--------------------------|-------------|
| Linux | `~/.config/gws/` | `~/.config/gws/` (correct) |
| macOS | `~/Library/Application Support/gws/` | `~/.config/gws/` (wrong) |
| Windows | `%APPDATA%\gws\` | `~/.config/gws/` (wrong) |

Users on macOS/Windows who followed the README and placed `client_secret.json` at `~/.config/gws/` had their config silently ignored because the CLI was reading from a different OS-specific path.

### Solution

- Changed `config_dir()` to use `~/.config/gws/` on all platforms (via `dirs::home_dir()`)
- Added backward-compatible fallback: if `~/.config/gws/` doesn't exist but the legacy OS-specific path does, it uses the legacy path (existing users are not affected)
- Consolidated duplicated `dirs::config_dir()` calls in `auth.rs` and `discovery.rs` to use the central `config_dir()` helper

### Fallback logic

1. `~/.config/gws/` exists → use it (new installs, Linux users)
2. `dirs::config_dir()/gws/` exists → use it (existing macOS/Windows users)
3. Neither exists → use `~/.config/gws/` (will be created on first write)

On Linux, both paths are identical so there is no behavioral change.

**Dry Run Output:**
```
N/A — config path resolution change, no HTTP request generated
```

## Checklist:

### DIFF
--- a/.changeset/fix-cross-platform-config-path.md
+++ b/.changeset/fix-cross-platform-config-path.md
@@ -1,0 +1,10 @@
+---
+"@googleworkspace/cli": patch
+---
+
+fix: use ~/.config/gws on all platforms for consistent config path
+
+Previously used `dirs::config_dir()` which resolves to different paths per OS
+(e.g. ~/Library/Application Support/gws on macOS, %APPDATA%\gws on Windows),
+contradicting the documented ~/.config/gws/ path. Now uses ~/.config/gws/
+everywhere with a fallback to the legacy OS-specific path for existing installs.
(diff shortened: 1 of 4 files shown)


jpoehnelt:
Can you run `cargo fmt`.

PR #164, opened 2026-03-05
fix: remove legacy darwin.apple_sdk reference in flake.nix

nklmilojevic:
## Description

`darwin.apple_sdk` has been removed from nixpkgs-unstable as a legacy compatibility stub. The Security framework is now provided by the default Apple SDK included in stdenv, so the explicit dependency is no longer needed.

Also updates `stdenv.isLinux`/`stdenv.isDarwin` to the non-deprecated `stdenv.hostPlatform.isLinux`/`stdenv.hostPlatform.isDarwin`.

## Checklist:

- [ ] My code follows the `AGENTS.md` guidelines (no generated `google-*` crates).
- [ ] I have run `cargo fmt --all` to format the code perfectly.
- [ ] I have run `cargo clippy -- -D warnings` and resolved all warnings.
- [ ] I have provided a Changeset file (e.g. via `pnpx changeset`) to document my changes.

^
N/A — this change only modifies flake.nix, no Rust/TS code was changed.

### DIFF
--- a/.changeset/fix-nix-flake-darwin.md
+++ b/.changeset/fix-nix-flake-darwin.md
@@ -1,0 +1,4 @@
+---
+---
+
+fix: remove legacy darwin.apple_sdk reference in flake.nix


malob:
Appreciate this, I ran into the same issue.

One additional suggestion: consider switching the flake's nixpkgs input from `nixos-unstable` to `nixpkgs-unstable`. `nixos-unstable` is gated on NixOS (Linux) CI, while `nixpkgs-unstable` includes Darwin CI checks — more appropriate for a tool that explicitly targets macOS. This would also reduce the chance of similar Darwin breakage slipping through in future lock updates.

jpoehnelt:
fixed in #154

PR #190, opened 2026-03-05
fix: log auth errors to stderr instead of silently discarding

Bortlesboat:
## Summary

When OAuth authentication fails in the main CLI flow, the error was silently discarded and the request proceeded unauthenticated. Users then received confusing 401/403 API errors with no indication of the root cause (expired token, missing credentials, wrong account, etc.).

Now prints the original error message and a hint to stderr.

## Changes

- `src/main.rs`: Replace `Err(_) => (None, AuthMethod::None)` with logging the error and a hint to run `gws auth login`

Note: The MCP server (`mcp_server.rs:430-435`) already logs auth failures correctly — this brings the main CLI in line with that pattern.

## Test plan

- [ ] CI: `cargo test`, `cargo clippy`, `cargo fmt`
- [ ] Manual: run any API command without auth configured, verify error + hint appear on stderr

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -240,7 +240,7 @@ async fn run() -> Result<(), GwsError> {
         Ok(t) => (Some(t), executor::AuthMethod::OAuth),
         Err(e) => {
             eprintln!("warning: authentication failed: {e:#}");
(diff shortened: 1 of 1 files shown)


jpoehnelt:
Seems obsolete.

PR #202, opened 2026-03-05
fix: store granted scopes in credentials, warn on scope escalation via export

### LINKED ISSUE #168: gws auth login --readonly + auth export --unmasked appears to allow full access on external machine
## Summary
When authenticating with read-only scopes using `gws auth login --readonly`, exporting credentials with:

```bash
gws auth export --unmasked > credentials.json
```

and then using those credentials on another machine (OpenClaw), the resulting access appears to still allow full/write operations.

## Reproduction
1. Run:
   ```bash
   gws auth login --readonly
   ```
2. Export credentials:
   ```bash
   gws auth export --unmasked > credentials.json
   ```
3. Move `credentials.json` to an external machine/environment (OpenClaw).
4. Use the exported credentials with `gws` there.
5. Attempt a write operation (for example, Gmail send).

## Actual behavior
Write/full-access behavior is still possible on the external machine, even though login was done with `--readonly`.

## Expected behavior
Read-only scope enforcement should persist with exported credentials and be enforced by Google-issued token scopes, not only by client-side command restrictions.

## Questions
- Are `--readonly` scopes currently enforced only in the local client flow, rather than in the OAuth token/scopes themselves?
- If so, what is the recommended way to ensure exported credentials are truly read-only on external machines?
- Can `--readonly` become the default behavior in the installer/setup flow (or at least a strongly guided default), not just a local/client-side option?

## Why this matters
Users exporting credentials for remote execution expect least-privilege guarantees to carry over across machines/environments.

efe-arv:
### Problem

When users authenticate with `--readonly` and export credentials via `gws auth export --unmasked`, the exported refresh token can be used to request **write access** on another machine. This happens because OAuth2 refresh tokens are not scope-limited by Google — the token can mint access tokens for **any** scope the OAuth app is authorized for.

### Changes

| File | Change |
|---|---|
| `src/auth_commands.rs` | `handle_login` now persists `granted_scopes` array in the encrypted credential JSON |
| `src/auth_commands.rs` | `handle_export` warns when credentials were obtained with readonly scopes (or lack scope metadata entirely) |
| `src/auth.rs` | `load_credentials_inner` reads `granted_scopes` and warns when a requested scope was not in the original grant |
| `src/auth_commands.rs` | 2 new tests: `credentials_json_includes_granted_scopes`, `credentials_without_granted_scopes_is_legacy` |

### How it works

**Login** (credential creation):
```json
{
  "type": "authorized_user",
  "client_id": "...",
  "client_secret": "...",
  "refresh_token": "...",
  "granted_scopes": ["https://...drive.readonly", "https://...gmail.readonly"]
}
```

**Export** (credential sharing):
```
⚠️  WARNING: These credentials were obtained with read-only scopes,
but the exported refresh token is NOT scope-limited by Google.
Anyone with this token can request write access.
Consider using a separate OAuth client for read-only use cases.
```

**Runtime** (scope enforcement):
```
⚠️  Scope 'https://...drive' was not in the granted scopes at login time.
Re-run 'gws auth login' with the required scopes.
```

### OAuth2 limitation

This fix **mitigates** but cannot fully **prevent** scope escalation. Google OAuth2 refresh tokens are fundamentally not scope-bound — the token can always request any scope the app was configured for. True scope isolation requires separate OAuth clients per access level. The export warning documents this limitation explicitly.

### Backward compatibility

- Credentials without `granted_scopes` (created before this fix) continue to work
- A separate warning is shown when exporting legacy credentials without scope metadata
- No breaking changes to the credential format (`granted_scopes` is an optional field)

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -280,6 +280,39 @@ async fn load_credentials_inner(
             .as_str()
             .ok_or_else(|| anyhow::anyhow!("Missing refresh_token in encrypted credentials"))?;
 
+        // Enforce scope boundaries: if credentials were obtained with specific scopes,
+        // reject requests for scopes outside the granted set.
+        // This prevents scope escalation when credentials are exported and reused.
+        if let Some(granted) = creds.get("granted_scopes").and_then(|v| v.as_array()) {
+            let granted_set: std::collections::HashSet<&str> = granted
+                .iter()
+                .filter_map(|s| s.as_str())
+                .collect();
+
+            if !granted_set.is_empty() {
+                for requested in scopes {
+                    if !granted_set.contains(requested) {
+                        // Check if a broader scope subsumes the requested one.
+                        // e.g. "drive" subsumes "drive.readonly"
+                        let prefix = "https://www.googleapis.com/auth/";
+                        let req_short = requested.strip_prefix(prefix).unwrap_or(requested);
+                        let is_subsumed = granted_set.iter().any(|&g| {
+                            let g_short = g.strip_prefix(prefix).unwrap_or(g);
+                            req_short.starts_with(g_short)
+                                && req_short.as_bytes().get(g_short.len()) == Some(&b'.')
+                        });
+                        if !is_subsumed {
+                            eprintln!(
+                                "⚠️  Scope '{}' was not in the granted scopes at login time. \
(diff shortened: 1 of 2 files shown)


jpoehnelt:
Missing CLA

PR #209, opened 2026-03-05
ci: auto-trigger Gemini Code Assist review on PR push

jpoehnelt:
Adds a `gemini-review` job to the automation workflow that posts `/gemini review` on every push to a PR.

**Rate limiting:** Uses GitHub Actions `concurrency` with `cancel-in-progress: true` and a 60-second debounce sleep. Rapid pushes within 60s only trigger one review.

**Trigger:** Only fires on `synchronize` (new pushes), not on `opened`/`reopened` since Gemini auto-reviews on initial PR creation.


jpoehnelt:
Going to try this as a test for the AI prs.

PR #301, opened 2026-03-08
fix(schema): keep repeated query metadata and honor array params

### LINKED ISSUE #300: `gws schema` drops `repeated: true` for array query parameters, causing metadata format to fail
## Summary

`gws schema` omits the `repeated: true` field for query parameters that accept multiple values. As a result, parameters like `metadataHeaders` appear as `string` instead of `string[]`, and passing an array via `--params` does not expand into repeated URL query parameters as expected.

## Environment

- `gws` version: 0.8.0

## Steps to Reproduce

```bash
gws schema gmail.users.messages.get
```

**`gws schema` output for `metadataHeaders`:**

```json
{
  "description": "When given and format is `METADATA`, only include headers specified.",
  "location": "query",
  "required": false,
  "type": "string"
}
```

**Google Discovery API raw response** (`https://www.googleapis.com/discovery/v1/apis/gmail/v1/rest`):

```json
{
  "location": "query",
  "repeated": true,
  "type": "string",
  "description": "When given and format is `METADATA`, only include headers specified."
}
```

The `repeated: true` field is present in the Discovery API but missing from `gws schema` output.

## Expected Behavior

1. `gws schema` should expose `repeated: true` so users know the parameter accepts multiple values.
2. When `repeated: true`, passing a JSON array via `--params` should expand into repeated URL query parameters:

   ```
   ?metadataHeaders=Subject&metadataHeaders=Date&metadataHeaders=From
   ```

## Actual Behavior

Passing an array has no effect:

```bash
# Does NOT return Subject/Date/From headers — payload.headers is absent
gws gmail users messages get --params '{"userId":"me","id":"<id>","format":"metadata","metadataHeaders":["Subject","Date","From"]}'
```

Passing a single string works, but only one header is returned:

```bash
# Works, but limited to one header
gws gmail users messages get --params '{"userId":"me","id":"<id>","format":"metadata","metadataHeaders":"Subject"}'
```

The workaround is to use `format: "full"`, which returns the entire message payload — significantly heavier than `metadata`.

## Impact

The `metadata` format exists specifically to fetch lightweight message metadata. This bug makes it impossible to retrieve multiple headers efficiently, forcing users to fall back to `format: "full"`.

sonwr:
This fixes #300.

`gws schema` now preserves `repeated: true` for method parameters, and request execution now expands JSON array values into repeated query keys for parameters marked as repeated in discovery docs.

Example behavior after this change:
- `metadataHeaders` appears with `repeated: true` in `gws schema gmail.users.messages.get`
- `--params '{"metadataHeaders":["Subject","Date","From"]}'` sends repeated query pairs

Verification:
- `~/.cargo/bin/cargo test test_param_to_json -- --nocapture`
- `~/.cargo/bin/cargo test test_build_url_ -- --nocapture`
- `~/.cargo/bin/cargo run -- schema gmail.users.messages.get` (confirmed `metadataHeaders.repeated=true` in output)

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -65,7 +65,7 @@ struct ExecutionInput {
     params: Map<String, Value>,
     body: Option<Value>,
     full_url: String,
-    query_params: HashMap<String, String>,
+    query_params: Vec<(String, String)>,
     is_upload: bool,
 }
 
@@ -484,7 +484,7 @@ fn build_url(
     method: &RestMethod,
     params: &Map<String, Value>,
     is_upload: bool,
-) -> Result<(String, HashMap<String, String>), GwsError> {
+) -> Result<(String, Vec<(String, String)>), GwsError> {
     // Build URL base and path
 
     // Actually we need to construct base URL properly if not present
@@ -521,7 +521,7 @@ fn build_url(
 
     // Substitute path parameters and separate query parameters
     let path_parameters = extract_template_path_parameters(path_template);
-    let mut query_params: HashMap<String, String> = HashMap::new();
+    let mut query_params: Vec<(String, String)> = Vec::new();
 
     for (key, value) in params {
         let val_str = match value {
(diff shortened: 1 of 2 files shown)


jpoehnelt:
can you add a changeset

PR #345, opened 2026-03-09
fix(credential_store): stop persisting encryption key file when keyring is available

### LINKED ISSUE #344: credential_store: .encryption_key file persists on disk even when OS keyring succeeds
## Bug

### Impact

### Root Cause

### Proposed Fix

```rust
// Generate key
let mut key = [0u8; 32];

// Try keyring first
} else {
}
```

### Verification

```bash
```

jpoehnelt:
## Summary

`get_or_create_key()` unconditionally wrote the encryption key to `~/.config/gws/.encryption_key` on first run, even when the OS keyring was available. This left key material on disk as a plain file (permissions `0600`), making credentials fully portable by copying the config directory.

## Changes

- **Extract `save_key_file()` helper** — deduplicates file-writing logic with secure permissions
- **Keyring success path**: delete stale `.encryption_key` if it exists (migrates existing installs on upgrade)
- **NoEntry + existing file**: migrate file-based key into keyring, then delete the file
- **NoEntry + new key**: try keyring first; only write `.encryption_key` as fallback if keyring store fails

## Migration

Existing users with a stale `.encryption_key` file will have it automatically removed on their next CLI invocation, as long as their OS keyring is working. No manual steps required.

## Testing

All 499 tests pass. The 8 `credential_store` tests specifically validate encrypt/decrypt round-trips and key determinism.

Fixes #344

### DIFF
--- a/.changeset/fix-encryption-key-file-leak.md
+++ b/.changeset/fix-encryption-key-file-leak.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Stop persisting encryption key to `.encryption_key` file when OS keyring is available. Existing file-based keys are migrated into the keyring and the file is removed on next CLI invocation.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,13 @@ jobs:
 
       - name: Enable sccache
         if: steps.sccache.outcome == 'success'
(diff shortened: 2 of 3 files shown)


jpoehnelt:
Addressed the review feedback from @gemini-code-assist: `save_key_file` now returns `std::io::Result<()>` and both call sites propagate errors via `?`. This prevents silent data loss if the fallback key file can't be written.

PR #357, opened 2026-03-10
fix(credential_store): preserve .encryption_key file for keyring-less environments

### LINKED ISSUE #344: credential_store: .encryption_key file persists on disk even when OS keyring succeeds

jpoehnelt:
## Summary

The fix for #344 introduced automatic deletion of `.encryption_key` when the OS keyring succeeds. In Docker/ephemeral-keyring environments, this permanently loses the encryption key on container restart, making `credentials.enc` undecryptable.

## Changes

- **Never delete `.encryption_key`** — the file persists as a safe fallback. The keyring remains authoritative when available.
- **Always save new keys to file**, even when keyring `set_password` succeeds. This ensures the key survives keyring loss (e.g. Docker container restart).
- **Extract `resolve_key()`** with a `KeyringProvider` trait for testability.
- **Add 13 new unit tests** covering all keyring/file scenarios (keyring ok, NoEntry, platform error, with/without file, set success/failure).

## Root Cause

The `NoEntry` + existing file path (line 113-114) migrated the file key into the keyring, then deleted the file. In Docker, `set_password()` can succeed within a session but the keyring doesn't persist across container restarts. On next start, both the keyring and the file are gone → key lost forever.

## Test Results

```
test result: ok. 21 passed; 0 failed; 0 ignored (credential_store tests)
test result: ok. 518 passed; 0 failed; 0 ignored (full suite)
```

Fixes #344 (regression)

### DIFF
--- a/skills/gws-gmail-forward/SKILL.md
+++ b/skills/gws-gmail-forward/SKILL.md
@@ -44,6 +44,7 @@ gws gmail +forward --message-id 18f1a2b3c4d --to dave@example.com --cc eve@examp
 ## Tips
 
 - Includes the original message with sender, date, subject, and recipients.
+- Sends the forward as a new message rather than forcing it into the original thread.
 
 ## See Also
 


jpoehnelt:
Superseded — switching to a gogcli-style backend selection pattern with env var control instead of removing the keyring entirely.

PR #359, opened 2026-03-10
feat(credential_store): add GOOGLE_WORKSPACE_CLI_KEYRING_BACKEND env var

### LINKED ISSUE #344: credential_store: .encryption_key file persists on disk even when OS keyring succeeds

jpoehnelt:
## Summary

Adds gogcli-style backend selection for encryption key storage, inspired by [gogcli](https://github.com/steipete/gogcli).

### `GOOGLE_WORKSPACE_CLI_KEYRING_BACKEND`

| Value | Behavior |
|---|---|
| `keyring` (default) | OS keyring with `.encryption_key` file fallback |
| `file` | `.encryption_key` file only — for Docker/CI/headless |

### Key changes

- **Never delete `.encryption_key`** — it always serves as a durable fallback for ephemeral keyring environments
- When generating new keys with `backend=keyring`, save to **both** keyring and file
- Extract `KeyringProvider` trait + `resolve_key()` for testability
- 25 unit tests covering both backends and all edge cases
- Updated env var help, README, AGENTS.md

### Root cause (#344)

The previous fix for #344 deleted `.encryption_key` when the OS keyring succeeded. In Docker containers, the keyring is ephemeral, so the key was permanently lost on container restart. This PR preserves the file and adds an explicit env var to bypass the keyring entirely.

Fixes #344

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -873,6 +873,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "yup-oauth2",
+ "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


olvgrolle:
@jpoehnelt - cargo.toml is missing options to actually activate the actual keyring saves on macos / linux / pc: 

proposed diff: 

--- a/Cargo.toml
  +++ b/Cargo.toml
  @@ -66,3 +66,12 @@
   [dev-dependencies]
   serial_test = "3.4.0"
   tempfile = "3"
  +
  +[target.'cfg(target_os = "macos")'.dependencies]
  +keyring = { version = "3.6.3", features = ["apple-native"] }
  +
  +[target.'cfg(target_os = "linux")'.dependencies]
  +keyring = { version = "3.6.3", features = ["linux-native"] }
  +
  +[target.'cfg(target_os = "windows")'.dependencies]
  +keyring = { version = "3.6.3", features = ["windows-native"] }

PR #370, opened 2026-03-10
fix(auth): verify keyring round-trip to prevent phantom key loss

mikeciesla:
## Summary

- After every `keyring::Entry::set_password()`, immediately call `get_password()` and compare — only trust the keyring if the round-trip succeeds
- Always persist the encryption key to `~/.config/gws/.encryption_key` as a file backup, even when the keyring appears to work
- Stop deleting the file backup when a keyring entry already exists (guards against future keyring failures)
- Switch `save_key_file` to atomic write (tmp + rename) for crash safety

## Problem

On macOS, `keyring::Entry::set_password()` can return `Ok(())` without actually persisting to Keychain (phantom write). This was introduced in #345 which removed the file fallback when the keyring "succeeded". The result: the encryption key is lost between runs and all commands fail with:

```
Decryption failed. Credentials may have been created on a different machine.
```

## Test plan

- [x] `cargo test` — 509 passed
- [x] Manual test on macOS Sequoia 15.7.4 with phantom keyring writes:
  - `gws auth login` generates key and saves to file
  - `gws drive files list` decrypts successfully across process restarts
  - Key fingerprint is consistent across runs (verified with debug tracing)

### DIFF
--- a/src/credential_store.rs
+++ b/src/credential_store.rs
@@ -22,8 +22,13 @@ use rand::RngCore;
 use std::sync::OnceLock;
 
 /// Persist the base64-encoded encryption key to a local file with restrictive
-/// permissions (0600 file, 0700 directory). Used only as a fallback when the OS
-/// keyring is unavailable.
+/// permissions (0600 file, 0700 directory).
+///
+/// The file is always kept as a backup even when the OS keyring is available.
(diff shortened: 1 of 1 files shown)


jpoehnelt:

PR #371, opened 2026-03-10
fix(auth): preserve Linux encryption key fallback

### LINKED ISSUE #364: Linux 0.9.1: encrypted credentials become undecryptable on same machine after key-storage change
## Summary
On Linux, `gws 0.9.1` can end up with unreadable encrypted OAuth credentials even though `credentials.enc` exists and the CLI previously authenticated successfully.

Current failure:

```json
{
  "error": {
    "code": 401,
    "message": "Authentication failed: Failed to decrypt credentials: Decryption failed. Credentials may have been created on a different machine. Run `gws auth logout` and `gws auth login` to re-authenticate.",
    "reason": "authError"
  }
}
```

And `gws auth status` reports:

- `encrypted_credentials_exists: true`
- `encryption_valid: false`
- `encryption_error: "Could not decrypt. May have been created on a different machine."`
- `plain_credentials_exists: false`

This looks like the same bug family as #27, but on Linux with `0.9.1` after the recent key-storage changes.

## Environment
- OS: Debian GNU/Linux 13 (headless-ish environment)
- `gws`: `0.9.1`
- `DBUS_SESSION_BUS_ADDRESS` is present
- `libsecret-1-0` is installed
- no visible `gnome-keyring-daemon` process
- `~/.config/gws/credentials.enc` exists
- `~/.config/gws/.encryption_key` does **not** exist

## Evidence
`gws auth status`:

```json
{
  "auth_method": "oauth2",
  "client_config": "/home/max/.config/gws/client_secret.json",
  "client_config_exists": true,
  "encrypted_credentials": "/home/max/.config/gws/credentials.enc",
  "encrypted_credentials_exists": true,
  "encryption_error": "Could not decrypt. May have been created on a different machine.",
  "encryption_valid": false,
  "plain_credentials": "/home/max/.config/gws/credentials.json",
  "plain_credentials_exists": false,
  "storage": "encrypted"
}
```

And an API call fails with:

```json
{
  "error": {
    "code": 401,
    "message": "Authentication failed: Failed to decrypt credentials: Decryption failed. Credentials may have been created on a different machine. Run `gws auth logout` and `gws auth login` to re-authenticate.",
    "reason": "authError"
  }
}
```

## Why this looks like a regression
Issue #27 was fixed by #28 (`fix(auth): stabilize encrypted credential key fallback`).

But `0.9.1` includes:

- `5872dbe: Stop persisting encryption key to .encryption_key file when OS keyring is available. Existing file-based keys are migrated into the keyring and the file is removed on next CLI invocation.`

On this Linux machine, there is no `.encryption_key` fallback file anymore, and the stored encrypted credentials are no longer decryptable. That suggests the CLI believed the OS keyring was stable/available at write time, but a later process could not retrieve the same key reliably.

## Suspected cause
The Linux keyring availability check may be too optimistic in some environments:
- write path succeeds via an OS-keyring code path,
- fallback `.encryption_key` is removed / never written,
- subsequent process cannot retrieve the same key,
- `credentials.enc` becomes effectively unreadable.

This may be specific to headless Linux / DBus / Secret Service combinations where keyring access is partially available but not stable across invocations.

## Expected
If the OS keyring is not reliably available on Linux, `gws` should preserve a stable fallback instead of leaving `credentials.enc` unreadable.

## Actual
`credentials.enc` exists but cannot be decrypted on the same machine, causing all authenticated commands to fail.

## Request
Could you confirm whether `5872dbe` introduced a Linux regression here, and whether Linux should retain `.encryption_key` fallback unless keyring retrieval is proven stable across fresh processes?

mkrueger12:
## Description

Fix a Linux credential storage regression in `v0.9.1` where encrypted OAuth credentials can become undecryptable on the same machine after the key-storage change in `5872dbe`.

This PR keeps a stable local `.encryption_key` fallback on Linux even when OS keyring operations appear to succeed.

Fixes #364.

**Dry Run Output:**
```json
// Not applicable: this PR does not add a new feature or command.
```

## Problem

Issue #364 reports a Linux environment where:

- `credentials.enc` exists
- `gws auth status` shows `encrypted_credentials_exists: true`
- `encryption_valid: false`
- authenticated commands fail with:
  - `Failed to decrypt credentials: Decryption failed. Credentials may have been created on a different machine`

The environment is a Debian/headless-ish setup with DBus present and `libsecret` installed, but without a clearly stable desktop keyring session.

## Root cause

Commit `5872dbe` changed the key persistence behavior to stop keeping `.encryption_key` once the OS keyring was available.

That is fine when the keyring is reliably available across fresh processes, but it is too optimistic for some Linux/Secret Service setups:

1. one process successfully writes the encryption key into the keyring
2. the CLI removes or does not keep the local `.encryption_key` fallback
3. a later process cannot retrieve the same key from Secret Service
4. `credentials.enc` is still present, but it can no longer be decrypted

In other words, the write path can succeed once while the read path is not stable across invocations.

### Linux behavior

On Linux, after a successful keyring read/write, keep `.encryption_key` synchronized as a stable fallback instead of removing it.

This applies to:

- reading an existing key from the keyring
- migrating a file-based key into the keyring
- generating a new key and successfully storing it in the keyring

### Non-Linux behavior

Keep the current behavior:

- if keyring access succeeds, the local fallback file is removed
- the keyring remains authoritative

## Why this approach

The regression appears to be Linux-specific and tied to environments where Secret Service is partially available but not reliably retrievable across processes.

Keeping the fallback on Linux is a conservative durability fix:

- it restores stable decryption across invocations
- it avoids making `credentials.enc` unreadable on the same machine
- it preserves the newer keyring integration instead of reverting it entirely

## Security tradeoff

This does slightly reduce defense-in-depth on Linux compared to keyring-only storage, because the AES key may also exist on disk in `.encryption_key`.

That said:

- the file is written with restrictive permissions (`0600`, parent dir `0700`)
- this does not introduce remote code execution or privilege escalation
- it is a local at-rest secret storage tradeoff in exchange for restoring reliable credential decryption

A stricter future approach could try to prove keyring stability across fresh processes before deleting the fallback. For this regression fix, preserving the fallback on Linux seemed like the least surprising and safest behavior.

## Tests and validation

Added coverage for the platform strategy in `credential_store`:

- `sync_key_file_after_keyring_success_matches_platform_strategy`
  - verifies that Linux persists the fallback key file
  - verifies that other platforms remove it
- `linux_strategy_persists_fallback_key_material`
  - verifies the Linux fallback file contains the expected key material

Also added the required changeset file:

- `.changeset/polite-eggs-burn.md`

Commands run locally:

```bash
cargo fmt --all
cargo test credential_store -- --nocapture
cargo clippy -- -D warnings
```

## Reproduction path validated during investigation

Checked out `v0.9.1` and traced the regression to the current logic in `src/credential_store.rs`:

- successful keyring access removes `.encryption_key`
- successful key generation + keyring store does not retain a file fallback

That matches the failure mode reported in #364.

## Checklist

### DIFF
--- a/.changeset/polite-eggs-burn.md
+++ b/.changeset/polite-eggs-burn.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Preserve the local `.encryption_key` fallback on Linux so encrypted OAuth credentials remain decryptable across fresh processes when Secret Service availability is unstable.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -713,6 +713,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "fs2"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9564fc758e15025b46aa6643b1b77d047d1a56a1aea6e01002ac0c7026876213"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "futures-channel"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -857,6 +867,7 @@ dependencies = [
  "derive_builder",
  "dirs",
  "dotenvy",
+ "fs2",
  "futures-util",
  "hostname",
  "keyring",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ anyhow = "1"
 clap = { version = "4", features = ["derive", "string"] }
 dirs = "5"
 dotenvy = "0.15"
+fs2 = "0.4"
 hostname = "0.4"
 reqwest = { version = "0.12", features = ["json", "stream", "rustls-tls-native-roots"], default-features = false }
 rand = "0.8"
--- a/src/credential_store.rs
+++ b/src/credential_store.rs
@@ -12,14 +12,61 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fs::File;
 use std::path::PathBuf;
 
 use aes_gcm::aead::{Aead, KeyInit, OsRng};
 use aes_gcm::{AeadCore, Aes256Gcm, Nonce};
 
+use fs2::FileExt;
 use keyring::Entry;
 use rand::RngCore;
 use std::sync::OnceLock;
+
+fn should_persist_fallback_key() -> bool {
+    cfg!(target_os = "linux")
+}
+
+fn persist_fallback_key(path: &std::path::Path, b64_key: &str) {
+    if let Err(e) = save_key_file(path, b64_key) {
+        eprintln!(
+            "Warning: failed to persist local fallback encryption key at {}: {e}",
+            path.display()
+        );
+    }
+}
+
+fn sync_key_file_after_keyring_success(path: &std::path::Path, b64_key: &str) {
+    if should_persist_fallback_key() {
+        persist_fallback_key(path, b64_key);
+    } else if path.exists() {
+        let _ = std::fs::remove_file(path);
+    }
+}
+
+fn acquire_key_creation_lock(config_dir: &std::path::Path) -> anyhow::Result<File> {
+    std::fs::create_dir_all(config_dir)?;
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        if let Err(e) = std::fs::set_permissions(config_dir, std::fs::Permissions::from_mode(0o700))
+        {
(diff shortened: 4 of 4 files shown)


mkrueger12:
Addressed the race condition comment.

What changed:
- added an inter-process lock around key creation/migration in get_or_create_key() using a config-dir lock file (`.encryption_key.lock`)
- re-check key resolution only after the lock is held, so concurrent fresh gws processes cannot generate different keys and overwrite each other
- kept the Linux fallback-file behavior from the original fix

Validation run locally after the update:
- `cargo fmt --all`
- `cargo test credential_store -- --nocapture`
- `cargo clippy -- -D warnings`

Latest commit on the branch: `de50550`

jpoehnelt:

Issue #375, opened 2026-03-10
Bug: Encryption key lost during keyring migration on macOS


jzumwalt:
## Summary

`get_or_create_key()` in `credential_store.rs` deletes the file-based encryption key after "migrating" it to the OS keyring, but the keyring entry is not readable by subsequent processes. This causes all encrypted credentials to become permanently undecryptable until the user discovers the workaround.

## Affected versions

- gws 0.9.1 (npm-installed binary, ad-hoc signed)
- macOS (Darwin 24.5.0, Apple Silicon)

## Symptoms

```
gws drive files list --params '{"pageSize": 10}'
{
  "error": {
    "code": 401,
    "message": "Authentication failed: Failed to decrypt credentials: Decryption failed. Credentials may have been created on a different machine. Run `gws auth logout` and `gws auth login` to re-authenticate.",
    "reason": "authError"
  }
}
```

- `gws auth login` succeeds, but any subsequent command fails with the above error
- `gws auth status` shows `encryption_valid: false`
- Running `gws auth logout` + `gws auth login` does **not** fix the issue (it recreates the same broken cycle)

## Root cause

In `src/credential_store.rs` lines 104-115, when the keyring has no entry but a `.encryption_key` file exists:

```rust
// If keyring is empty, prefer a persisted local key first.
if key_file.exists() {
    if let Ok(b64_key) = std::fs::read_to_string(&key_file) {
        if let Ok(decoded) = STANDARD.decode(b64_key.trim()) {
            if decoded.len() == 32 {
                let mut arr = [0u8; 32];
                arr.copy_from_slice(&decoded);
                // Migrate file key into keyring; remove the
                // file if the keyring store succeeds.
                if entry.set_password(b64_key.trim()).is_ok() {
                    let _ = std::fs::remove_file(&key_file);  // <-- deletes the only working copy
                }
                return Ok(cache_key(arr));
            }
        }
    }
}
```

The problem:

1. `entry.set_password()` returns `Ok(())` — the write appears to succeed
2. The file is deleted because `set_password` returned `Ok`
3. The key lives in the `OnceLock` for the remainder of this process (login works)
4. On the **next** process invocation, `entry.get_password()` fails to find the keyring entry (likely due to macOS Security framework ACL restrictions on ad-hoc signed binaries)
5. The file is gone, so a **new** random key is generated
6. The new key cannot decrypt the existing `credentials.enc`

This creates an unrecoverable loop: every login encrypts with a key that is lost when the process exits.

[...]

Issue #367, opened 2026-03-10
macOS Keychain write silently fails: credentials.enc written but decryption key never saved (v0.9.1)


kwpark22:
**Environment**
- gws version: 0.9.1
- OS: macOS (Apple Silicon)
- Installed via: `npm install -g @googleworkspace/cli` (without sudo, after fixing `/usr/local` permissions)

**Problem**

After a successful `gws auth login` (returns success JSON with `credentials.enc` saved), all API calls fail with:

```json
{
  "error": {
    "code": 401,
    "message": "Authentication failed: Failed to decrypt credentials: Decryption failed. Credentials may have been created on a different machine.",
    "reason": "authError"
  }
}
```

**Root Cause**

The encryption key is **never saved to macOS Keychain**. Verified with:

```bash
$ security find-generic-password -s "gws" 2>&1
security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
```

`credentials.enc` is written successfully, but the decryption key doesn't exist in Keychain — making it permanently unreadable. No error or warning is shown during `gws auth login`.

**Steps to Reproduce**

1. `npm install -g @googleworkspace/cli` (without sudo)
2. `rm -rf ~/.config/gws/`
3. `gws auth setup` → success
4. `gws auth login` → OAuth completes, returns success JSON with `credentials.enc`
5. `security find-generic-password -s "gws"` → **not found**
6. `gws drive files list --params '{"pageSize": 5}'` → 401 decryption failed

**Additional Context**

Initially installed with `sudo npm install -g`, which may have triggered the first occurrence (sudo runs as root, which cannot write to the user Keychain). After reinstalling without sudo and doing a full `rm -rf ~/.config/gws/`, the bug persists — Keychain write still silently fails with no error output.

`gws auth export --unmasked` also fails (outputs the error JSON instead of credentials), confirming the key is truly missing from Keychain.

**Suggested Fix**

1. Log an explicit error/warning when Keychain write fails
2. Automatically fall back to `.encryption_key` file if Keychain is unavailable
3. Consider making a `--no-keyring` flag available as a workaround option

Related: #151

annikovk:
See https://github.com/googleworkspace/cli/issues/360

jpoehnelt:
[...]

Issue #274, opened 2026-03-06
[Bug] gws auth login fails on macOS - Keychain integration not working


nadia1971:
---
name: 🐛 Bug Report
about: Create a report to help us improve
title: "[Bug] gws auth login fails on macOS - Keychain integration not working, credentials not saved"
labels: bug
---

## Bug Description

On macOS, `gws auth login` completes the OAuth flow successfully, but credentials are not properly saved or recognized. Subsequent commands fail with "Access denied. No credentials provided."

## Environment

- **OS**: macOS (Darwin 25.3.0, arm64)
- **gws version**: 0.4.4
- **Installation**: Homebrew (`npm install -g @googleworkspace/cli`)
- **Node version**: v25.6.1

## Steps to Reproduce

1. Run `gws auth login`
2. Complete OAuth authentication in browser
3. See "Authentication successful. Encrypted credentials saved." message
4. Run any gws command (e.g., `gws drive files list`)
5. Get error: "Access denied. No credentials provided"

## Expected Behavior

After successful authentication, credentials should be saved and subsequent commands should work without re-authentication.

## Actual Behavior

```bash
$ gws auth login
Open this URL in your browser to authenticate:
  https://accounts.google.com/o/oauth2/auth?...

# Complete authentication in browser

{
  "account": "(unknown)",
  "credentials_file": "/Users/yangtian/.config/gws/credentials.enc",
  "encryption": "AES-256-GCM (key secured by OS Keyring or local `.encryption_key`)",
  "message": "Authentication successful. Encrypted credentials saved.",
  "status": "success"
}

$ gws drive files list --params '{"pageSize": 3}'
{
  "error": {
    "code": 401,
    "message": "Access denied. No credentials provided. Run `gws auth login`...",
    "reason": "authError"
  }
}

$ gws auth list
{
  "accounts": [],
  "count": 0,
  "default": ""
}
```

### Files Created

```
~/.config/gws/
├── .encryption_key        # ✅ Created
├── credentials.enc        # ✅ Created (encrypted)
└── client_secret.json     # ✅ Exists
```

### Keychain Status

```bash
# No gws-cli entry in macOS Keychain
$ security find-generic-password -s "gws-cli" -a "google-oauth" -w
security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
```

### Workaround

Python-based authentication works correctly:
- Using `google-auth` library directly
- Token saved to `~/.gws/token.json`
- Commands work when using Python scripts

[...]

PR #379, opened 2026-03-10
fix(auth): verify keyring roundtrip before deleting .encryption_key

### LINKED ISSUE #360: macOS: Credential decryption always fails - OS Keyring not storing encryption key
## Description

`gws auth login` completes successfully and reports encrypted credentials saved, but all subsequent commands fail with:

```
Authentication failed: Failed to decrypt credentials: Decryption failed. Credentials may have been created on a different machine.
```

This happens consistently, including when running `gws auth login` and subsequent commands from the **same terminal session**.

## Environment

- macOS Sequoia (Darwin 25.3.0)
- gws 0.9.1
- Apple Silicon (M-series)
- zsh shell

## Steps to Reproduce

1. `gws auth setup` (or manually configure OAuth client)
2. `gws auth login` → completes successfully, browser OAuth flow works
3. `gws drive files list --params '{"pageSize": 3}'` → fails with decryption error
4. `gws auth export` → also fails with same error

## Investigation

- No keychain entries found for `gws` via `security find-generic-password` or `security dump-keychain`
- The Rust keyring crate appears to not be writing to macOS Keychain
- Created `.encryption_key` file in `~/.config/gws/` as mentioned in the status output ("OS Keyring or local `.encryption_key`") — still fails
- `GWS_STORAGE=plain` env var has no effect
- `gws auth status` confirms `encryption_valid: false`

## Expected Behavior

Either:
1. OS Keyring integration should work on macOS, or
2. The `.encryption_key` fallback should be functional, or
3. A `--no-encrypt` / plain storage option should be available

## Workaround

None found. The tool is currently unusable on macOS.

alexriftagentic:
## Summary

Fixes the credential decryption bug that makes `gws` unusable on macOS (and some Linux environments) after `gws auth login`.

- **Root cause**: `get_or_create_key()` in `credential_store.rs` deletes the `.encryption_key` file after `entry.set_password()` returns `Ok(())`. On macOS with ad-hoc signed binaries (e.g. npm-installed), the Keychain write silently fails to persist — so the next process invocation can't read the key back, generates a new random key, and existing `credentials.enc` becomes permanently undecryptable.
- **Symptom**: `"encryption_error": "Could not decrypt. May have been created on a different machine."` immediately after a successful `gws auth login`, on the same machine.

## Changes

1. **Migration path** (existing file key → keyring): After `set_password()`, verify the roundtrip with `get_password()`. Only delete `.encryption_key` if readback matches.
2. **New key generation**: Always write to the local `.encryption_key` file first (source of truth), then optionally store in keyring. Previously it wrote to keyring only and skipped the file if `set_password()` returned `Ok`.
3. **Keyring-read-success path**: Stop proactively deleting the `.encryption_key` file when keyring read succeeds — keep it as a durable backup for environments where keyring access is intermittent.

## Test plan

- [x] `cargo check` passes
- [x] On macOS (Apple Silicon, npm install): `gws auth login` → `gws auth status` shows `encryption_valid: true`
- [x] On macOS: `.encryption_key` file is preserved after login
- [ ] On Linux with no keyring daemon: same flow works via file fallback
- [ ] Existing unit tests pass (`encrypt_decrypt_round_trip`, etc.)

## Fixes

Closes #360, #364, #367, #375, #274

### DIFF
--- a/src/credential_store.rs
+++ b/src/credential_store.rs
@@ -89,11 +89,9 @@ fn get_or_create_key() -> anyhow::Result<[u8; 32]> {
                     if decoded.len() == 32 {
                         let mut arr = [0u8; 32];
                         arr.copy_from_slice(&decoded);
-                        // Keyring is authoritative — remove redundant file copy
-                        // if it exists (migrates existing installs on upgrade).
-                        if key_file.exists() {
-                            let _ = std::fs::remove_file(&key_file);
-                        }
+                        // Keyring read succeeded — keep the file as a backup
+                        // in case a future process cannot access the keyring
+                        // (e.g. ad-hoc signed binaries on macOS).
                         return Ok(cache_key(arr));
                     }
                 }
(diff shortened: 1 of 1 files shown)


jpoehnelt:

PR #396, opened 2026-03-11
fix: refresh OAuth2 tokens in long-running watch/subscribe loops

### LINKED ISSUE #392: Bug: `gmail +watch` and `events +subscribe` exit after ~1 hour due to expired access token
### Description

The long-running helper commands `gmail +watch` and `events +subscribe` obtain OAuth2 access tokens once at startup and pass them as static `&str` values into their pull loops. Since Google access tokens expire after 3600 seconds (1 hour), the Pub/Sub pull request eventually receives a 401 response, which the loop treats as a fatal error and exits.

### Steps to reproduce

```sh
gws gmail +watch \
  --subscription projects/my-project/subscriptions/my-sub \
  --label-ids INBOX \
  --poll-interval 5
```

Wait ~60 minutes. The process exits with status 1.

### Expected behavior

The `+watch` command should run indefinitely (up to the 7-day Gmail watch expiration), automatically refreshing the access token before or when it expires.

### Actual behavior

The process exits with a non-zero status after ~1 hour on every run. The lifetime is consistent across restarts, matching the Google OAuth2 access token TTL of 3600 seconds.

### Root cause

In `src/helpers/gmail/watch.rs`, tokens are obtained once before the loop:

```rust
// watch.rs lines 17-22
let gmail_token = auth::get_token(&[GMAIL_SCOPE]).await …;
let pubsub_token = auth::get_token(&[PUBSUB_SCOPE]).await …;
```

`auth::get_token()` (in `src/auth.rs` line 91) builds a `yup_oauth2` authenticator, calls `.token()` once to get the access token as a `String`, then drops the authenticator, discarding its refresh capability.

The tokens are passed into `watch_pull_loop()` as `&str`:

```rust
// watch.rs lines 247-254
async fn watch_pull_loop(
    client: &reqwest::Client,
    pubsub_token: &str,
    gmail_token: &str,
    …
```

When the token expires after 3600s, the next Pub/Sub pull returns a 401, which hits the fatal error path at line 282-289:

```rust
if !resp.status().is_success() {
    let body = resp.text().await.unwrap_or_default();
    return Err(GwsError::Api { … });
}
```

The same pattern exists in `src/helpers/events/subscribe.rs` (`pull_loop` takes `token: &str`).

The HTTP client (`src/client.rs`) is a plain `reqwest::Client` with no auth middleware.

### Suggested fix

Retain the `yup_oauth2` authenticator and call `.token()` before each API request (or on a timer), rather than extracting the access token string once.

Alternatively, `get_token()` could be called periodically inside the loop (e.g., every 45 minutes or upon receiving a 401).

### Affected commands

- `gws gmail +watch` (`src/helpers/gmail/watch.rs`)
- `gws events +subscribe` (`src/helpers/events/subscribe.rs`)
- Potentially any future long-running helper that calls `auth::get_token()` once

### Version

gws 0.8.0

### Environment

Linux

zerone0x:
Fixes #392

## Summary
- Refresh OAuth2 tokens inside the pull loops of `gmail +watch` and `events +subscribe` by calling `auth::get_token()` on each iteration, instead of obtaining the token once at startup
- Also refresh tokens before cleanup operations that run after the long-lived loop exits
- `get_token()` uses cached credentials via `yup_oauth2` and only contacts the OAuth server when the cached access token has expired, so this adds negligible overhead

## Changes
- `src/helpers/gmail/watch.rs`: Removed `pubsub_token` and `gmail_token` parameters from `watch_pull_loop`; tokens are now refreshed inside the loop. Cleanup section also refreshes its token.
- `src/helpers/events/subscribe.rs`: Removed `token` parameter from `pull_loop`; token is now refreshed inside the loop. Cleanup section also refreshes its token.

## Test plan
- [x] `cargo check` passes
- [x] `cargo test` — all 550 tests pass
- [x] `cargo clippy` — no warnings
- [ ] Manual: run `gws gmail +watch` for >1 hour and verify it does not exit with 401
- [ ] Manual: run `gws events +subscribe` for >1 hour and verify it does not exit with 401

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/helpers/events/subscribe.rs
+++ b/src/helpers/events/subscribe.rs
@@ -248,25 +248,30 @@ pub(super) async fn handle_subscribe(
         };
 
     // Pull loop
-    let result = pull_loop(&client, &pubsub_token, &pubsub_subscription, config.clone()).await;
+    let result = pull_loop(&client, &pubsub_subscription, config.clone()).await;
 
     // On exit, print reconnection info or cleanup
     if created_resources {
         if config.cleanup {
             eprintln!("\nCleaning up Pub/Sub resources...");
+            // Refresh the token for cleanup — the original may have expired
+            // during the long-running pull loop.
+            let cleanup_token = auth::get_token(&[PUBSUB_SCOPE])
+                .await
+                .unwrap_or_default();
             // Delete Pub/Sub subscription
             let _ = client
                 .delete(format!(
                     "https://pubsub.googleapis.com/v1/{pubsub_subscription}"
                 ))
-                .bearer_auth(&pubsub_token)
+                .bearer_auth(&cleanup_token)
                 .send()
                 .await;
             // Delete Pub/Sub topic
             if let Some(ref topic) = topic_name {
                 let _ = client
                     .delete(format!("https://pubsub.googleapis.com/v1/{topic}"))
-                    .bearer_auth(&pubsub_token)
+                    .bearer_auth(&cleanup_token)
                     .send()
                     .await;
             }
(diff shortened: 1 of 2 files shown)


etanase:
Didn't see this before opening #407 

This fix is simpler but doesn't include regression testing. #407 adds an abstraction to the authentication to allow for testing, but is indeed more complex.

jpoehnelt:
see #407

PR #398, opened 2026-03-11
fix(auth): auto-recover from credential decryption failures after upgrade

### LINKED ISSUE #389: Auth broken after update
I upgraded to .11 after initially authing with a pre-keyring version. I ran `gws auth logout` and then `gws auth login` and successfully authenticated. I get this error when running the Drive list files command, though:

```
~> gws drive files list --params '{"pageSize": 10}'
{
  "error": {
    "code": 401,
    "message": "Authentication failed: Failed to decrypt credentials: Decryption failed. Credentials may have been created on a different machine. Run `gws auth logout` and `gws auth login` to re-authenticate.",
    "reason": "authError"
  }
}
```

I tried a fresh reinstall (deleting everything but client_secrets) and got the same error

zerone0x:
## Summary

Fixes #389

After upgrading from a pre-keyring version to v0.11+, users encounter `Failed to decrypt credentials: Decryption failed` errors that persist even after `logout + login` cycles. The root cause is that the encryption key can change between versions (the keyring migration in #345 could delete the `.encryption_key` file while the keyring entry might be lost later), leaving `credentials.enc` encrypted with a now-unavailable key.

**Changes:**

- **`auth.rs`**: When `credentials.enc` fails to decrypt, automatically remove the stale file and associated token caches, then return a clear error directing the user to `gws auth login`. This breaks the "stuck after logout+login" cycle by ensuring stale credentials don't block recovery.
- **`credential_store.rs`**: When the OS keyring returns a valid encryption key but no `.encryption_key` backup file exists, create one. This prevents future key loss if the keyring becomes unavailable (container restart, daemon change, OS upgrade).

## Root Cause Analysis

1. Pre-keyring versions stored the encryption key only in `~/.config/gws/.encryption_key`
2. The v0.11 keyring migration (#345) moved the key to the OS keyring and deleted the file
3. If the keyring later becomes unavailable (flaky D-Bus, different session, etc.), a new key is generated
4. Existing `credentials.enc` is encrypted with the old (now lost) key and cannot be decrypted
5. The old code returned a hard error with no automatic recovery, leaving users stuck

## Test plan

- [x] `cargo check` passes
- [x] `cargo clippy` passes with no warnings
- [x] `cargo test` — all 552 tests pass
- [x] New test: `test_load_credentials_corrupt_encrypted_file_is_removed` — verifies stale credentials are cleaned up on decryption failure
- [x] New test: `keyring_backend_creates_file_backup_when_missing` — verifies file backup is created when keyring has key but file is missing
- [ ] Manual test: simulate upgrade scenario (encrypt with one key, switch key, verify auto-recovery)

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -207,8 +207,32 @@ async fn load_credentials_inner(
 
     // 2. Encrypted credentials (always AuthorizedUser for now)
     if enc_path.exists() {
-        let json_str = credential_store::load_encrypted_from_path(enc_path)
-            .context("Failed to decrypt credentials")?;
+        let json_str = match credential_store::load_encrypted_from_path(enc_path) {
+            Ok(s) => s,
+            Err(e) => {
+                // Decryption failed — the encryption key likely changed (e.g. after
+                // an upgrade that migrated keys between keyring and file storage).
+                // Remove the stale file so the next `gws auth login` starts fresh,
+                // and fall through to other credential sources.
+                eprintln!(
+                    "warning: removing undecryptable credentials file ({}): {e:#}",
+                    enc_path.display()
+                );
+                let _ = std::fs::remove_file(enc_path);
+                // Also remove any stale token cache that used the old key.
+                let token_cache = enc_path.with_file_name("token_cache.json");
+                let _ = std::fs::remove_file(&token_cache);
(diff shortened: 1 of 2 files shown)


jpoehnelt:
see #435

PR #418, opened 2026-03-11
fix: stream multipart uploads to avoid OOM on large files

### LINKED ISSUE #244: Bug: Out-Of-Memory (OOM) Crash on Large File Uploads (Google Drive/YouTube)
## Description

The `gws` CLI supports uploading files (to Google Drive) via the `--upload` flag. However, the current implementation reads the **entire file into memory twice** before sending the HTTP request, and due to Rust's vector reallocation strategy, actually allocates up to **4x the file size in memory**. For large files (video files or large archives), this causes an immediate memory exhaustion (OOM) crash.

In `src/executor.rs` (`build_http_request`), the CLI does:

```rust
// 1. Reads the ENTIRE file into a Vec<u8> in memory
let file_bytes = tokio::fs::read(upload_path).await.map_err(...)?;

// 2. Passes the entire byte slice to build_multipart_body
let (multipart_body, content_type) = build_multipart_body(&input.body, &file_bytes)?;
```

Then in `build_multipart_body`:

```rust
let mut body = Vec::new(); // Starts empty, grows dynamically 
// ... appends boundaries and metadata ...
// 3. Allocates a SECOND copy of the entire file in memory
body.extend_from_slice(file_bytes); 
```

### Proof (Mock Test)
We wrote a memory-tracking allocator test mimicking the exact logic from `executor.rs` on a 500 MB mock file vs a streaming approach. Here are the true memory allocation results from the OS level:

**Current Implementation Memory:**
```text
=== OOM VULNERABILITY MOCK TEST ===
Simulating an upload of a 500 MB file...
[Step 1] tokio::fs::read completed. Memory allocated: 500 MB
[Step 2] build_multipart_body completed. Memory allocated: 1500 MB

Total memory allocated: 2000 MB (4.00x Overhead)
```

**Proposed Streaming Fix Memory:**
```text
=== STREAMING FIX MOCK TEST ===
Simulating an upload of a 500 MB file via chunks...

Total memory allocated: 0.06 MB (0.0001x Overhead)
```

Because `body` is initialized as an empty `Vec::new()` and `extend_from_slice` is called with a massive 500MB slice, Rust's vector capacity doubling strategy forces it to allocate far more memory than strictly necessary while growing the vector. Uploading a **5 GB file asks the OS for 20 GB of contiguous RAM**, guaranteeing a crash.

## Steps to Reproduce

1. Create a large dummy file (example 5 GB):
   ```bash
   dd if=/dev/zero of=large_video.mp4 bs=1G count=5
   ```
2. Attempt to upload it to Google Drive using `gws`:
   ```bash
   gws drive files create --upload large_video.mp4 --params '{"name": "large_video.mp4"}'
   ```
3. Observe the CLI crash or get killed by the OS OOM killer as memory usage spikes.

## Expected Behavior

The CLI should stream the file chunks directly from the disk to the network socket, maintaining a small, constant memory footprint regardless of the file size. As proven in the mock test, this requires `< 1MB` of RAM overhead.

## Suggested Fix

Since Google APIs require `multipart/related` (which `reqwest::multipart::Form` does not natively support, as it only does `multipart/form-data`), construct an async stream and pass it to `reqwest::Body::wrap_stream`. 

The stream should yield:
1. `Bytes` containing the opening boundary and JSON metadata part.
2. File chunks yielded via `tokio_util::codec::FramedRead` (reading the file sequentially).
3. `Bytes` containing the closing boundary.

This guarantees O(1) memory usage (a few MBs for the chunk buffer) during uploads.

Acelogic:
## Summary

Fixes #244 — uploading large files via `--upload` causes an out-of-memory crash because the entire file is read into memory twice (once by `tokio::fs::read`, then copied into a second `Vec` by `build_multipart_body`). A 5 GB file requests ~20 GB of contiguous RAM.

This replaces the buffered approach with a streaming `multipart/related` body:

- **`build_multipart_stream`** yields the body in three phases via `futures_util::stream::unfold`: preamble (boundary + JSON metadata), file chunks (64 KB reads from disk), and postamble (closing boundary)
- `Content-Length` is computed from `tokio::fs::metadata` so Google APIs still receive the correct header without buffering the file
- Memory usage is now O(1) (~64 KB buffer) regardless of file size

The old `build_multipart_body` is retained under `#[cfg(test)]` for the existing unit tests.

## Test plan

- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo test` — 552/552 pass (2 new tests added)
- [x] `test_build_multipart_stream_content_length` — verifies declared Content-Length matches preamble + file size + postamble
- [x] `test_build_multipart_stream_large_file` — 256 KB file (larger than 64 KB chunk size) verifies multi-chunk streaming arithmetic

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -851,6 +851,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "base64",
+ "bytes",
  "chrono",
  "clap",
  "crossterm",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ async-trait = "0.1.89"
 serde_yaml = "0.9.34"
 percent-encoding = "2.3.2"
 zeroize = { version = "1.8.2", features = ["derive"] }
+bytes = "1.11.1"
 
 
 # The profile that 'cargo dist' will build with
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -811,13 +811,14 @@ fn build_multipart_stream(
     let content_type = format!("multipart/related; boundary={boundary}");
 
     // Chain: preamble bytes -> file chunks (via ReaderStream) -> postamble bytes
+    // All parts use bytes::Bytes for zero-copy streaming.
     let file_path = file_path.to_owned();
-    let preamble_bytes = preamble.into_bytes();
-    let postamble_bytes = postamble.into_bytes();
+    let preamble_bytes = bytes::Bytes::from(preamble.into_bytes());
+    let postamble_bytes = bytes::Bytes::from(postamble.into_bytes());
 
     let file_stream =
         futures_util::stream::once(async move { tokio::fs::File::open(file_path).await })
-            .map_ok(|f| tokio_util::io::ReaderStream::new(f).map_ok(|b| b.to_vec()))
+            .map_ok(tokio_util::io::ReaderStream::new)
             .try_flatten();
 
     let stream =


Acelogic:
I]have signed the Google CLA. Please re-check.

PR #459, opened 2026-03-13
chore: add package lockfile

rotprods:
Adds the generated lockfile committed in this chat session.


rotprods:
Thanks for flagging this. You are correct: this repository uses pnpm. I am closing this PR because package-lock.json should not be added here.

PR #395, opened 2026-03-11
feat(gmail): add --attachment flag to +send, +reply, +reply-all, +forward

pae23:
## Summary

Adds file attachment support to Gmail helper commands (`+send`, `+reply`, `+reply-all`, `+forward`).

Currently, sending emails with attachments requires manually constructing a base64-encoded MIME message and passing it via `--json '{"raw": "..."}'`, which hits the Linux execve argument size limit (~128KB) for any non-trivial attachment. This PR adds a simple `--attachment` flag that handles MIME construction automatically.

## Changes

- **New `--attachment` flag** on `+send`, `+reply`, `+reply-all`, `+forward` — uses `ArgAction::Append` so each file gets its own `--attachment` flag (avoids comma-parsing issues with filenames)
- **`Attachment` struct** and `read_attachments()` helper — accepts both relative and absolute paths
- **`build_with_attachments()` method** on `MessageBuilder` — builds multipart/mixed MIME with shared `build_headers()`, random boundary, RFC 2045 base64 line-folding, content-type auto-detection (25+ types)
- **Filename security**: basename only (no path leakage), quote escaping (RFC 2045/2822), RFC 2231 non-ASCII encoding
- **Backward compatible** — delegates to `build()` when no attachments
- **No new dependencies** — uses existing `rand` and `base64` crates

## Usage

```bash
gws gmail +send --to alice@example.com --subject Report --body "See attached" --attachment report.pdf
gws gmail +send --to alice@example.com --subject Files --body "Two files" --attachment a.pdf --attachment b.csv
gws gmail +reply --message-id 18f1a2b3c4d --body "Updated version" --attachment updated.docx
gws gmail +forward --message-id 18f1a2b3c4d --to bob@example.com --attachment notes.pdf
```

## Test plan

- [x] `cargo build` passes
- [x] `cargo test` — all 147 gmail tests pass (10 new tests added)
- [x] Manual test: send email with ZIP attachment via `+send`

Generated with [Claude Code](https://claude.com/claude-code)

Issue #247, opened 2026-03-06
Gmail: cannot send emails with attachments (--json arg limit, --upload wrong content type)


0xBigBoss:
## Problem

There's no way to send a Gmail message with an attachment using `gws`. Two approaches fail:

### 1. `--json` with base64url-encoded raw MIME — hits OS argument length limit

The documented approach for sending raw MIME (with `threadId`, `In-Reply-To`, `References` headers, and attachments) is to base64url-encode the entire message and pass it via `--json '{"raw": "..."}'`. But even a modest ~577KB PDF attachment produces a ~1MB base64 payload, which exceeds macOS's ~256KB `ARG_MAX` for command-line arguments:

```
gws-behav gmail users messages send --params '{"userId": "me"}' --json "{\"threadId\": \"...\", \"raw\": \"$RAW\"}"
# => zsh: argument list too long: gws
```

### 2. `--upload` with raw MIME file — wrong content type

Gmail's `/upload/gmail/v1/users/me/messages/send` endpoint accepts `message/rfc822` for media uploads. But `--upload` wraps the file in `multipart/related` with an auto-detected content type, which Gmail rejects:

```
gws gmail users messages send --params '{"userId": "me", "uploadType": "media"}' --upload message.eml
# => Media type 'multipart/related; boundary=gws_boundary_...' is not supported.
```

## Suggested fixes

Any of these would unblock the use case:

1. **`--json-file <PATH>`** or **`--json @<PATH>`** — read JSON body from a file (like `curl -d @file.json`)
2. **`--json -`** — read JSON body from stdin
3. **`--upload-type <MIME>`** — let users override the content type for `--upload` (e.g. `--upload-type message/rfc822`)
4. **Auto-detect `.eml` → `message/rfc822`** in `--upload` content type detection

Options 1 or 2 are the most general and would help any API method with large request bodies, not just Gmail.

## Environment

- gws 0.4.4
- macOS Darwin 25.3.0 (arm64)

## Related

- #88 — first-class reply/forward support would also solve this at a higher level
- #244 — OOM on large file uploads (related large-payload issue)

chrisvariety:
workaround @ https://github.com/googleworkspace/cli/issues/276#issuecomment-4043198290 for `--json` to accept `@file`

PR #491, opened 2026-03-15
refactor(gmail): replace hand-rolled email construction with mail-builder

malob:
## Description

Replace hand-rolled MessageBuilder, RFC 2047 encoding, and header sanitization with the `mail-builder` crate (Stalwart Labs, 24KB, 1 optional runtime dep), and add `--from` flag to `+send` for send-as alias consistency with `+reply`, `+reply-all`, and `+forward`.

### Why mail-builder

The Gmail helpers have grown from simple text emails to supporting HTML mode, CC/BCC, reply-all with recipient dedup, forwarding with threading, and send-as aliases. Each feature added more hand-rolled RFC 5322 logic: MIME headers, content-type selection, RFC 2047 encoding for non-ASCII names (#482), header injection prevention, and address parsing/formatting.

Adding attachment support (#247) would require multipart/mixed MIME construction — boundaries, part encoding, content-disposition headers — on top of an already complex custom implementation. `mail-builder` gives us correct RFC 2047 encoding, Content-Transfer-Encoding, and MIME structure for free, plus a clean path to `--attach` via native multipart support.

### Commit 1 — Refactor

**Types:** `Mailbox` (parsed display name + email) replaces raw string passing. `OriginalMessage` fields use `Option<T>` instead of empty-string sentinels. Config types use `Vec<Mailbox>`. Message IDs stored bare (no angle brackets), parsed once at the boundary.

**Message construction:** Each command builds a `mail_builder::MessageBuilder` directly via shared helpers (`apply_optional_headers`, `set_threading_headers`, `finalize_message`). The pattern is consistent across all three commands.

**Security:** `sanitize_control_chars` in `Mailbox::parse` strips ASCII control characters (CRLF, null, tab) at the parse boundary. This supersedes `sanitize_header_value`, `sanitize_component`, and `encode_address_header` from #482 — mail-builder's structured address types prevent header injection structurally, and parse-boundary sanitization provides defense-in-depth. End-to-end injection tests verify CRLF in `--from`/`--cc` cannot create spurious headers.

**Behavioral changes:**
- Messages missing optional headers (Date, Cc) now cleanly omit those lines rather than rendering empty placeholders
- `parse_original_message` now returns `Result` — rejects messages missing required headers (threadId, From, Message-ID) instead of silently proceeding with empty fields
- mail-builder automatically adds Date and Message-ID headers (Gmail ignores both)
- Content-Transfer-Encoding is now properly set

### Commit 2 — `--from` on `+send`

Adds the `--from` flag to `+send`, consistent with `+reply`, `+reply-all`, and `+forward`. Uses the same `parse_optional_mailboxes` path and `apply_optional_headers` plumbing.

### Note on #482

This PR supersedes the RFC 2047 address header encoding merged in #482. mail-builder handles RFC 2047 automatically via structured `Address` types, and `sanitize_control_chars` in `Mailbox::parse` strips all ASCII control characters at the parse boundary (covering the same CRLF, null, and tab injection vectors as `sanitize_component`/`sanitize_header_value`). One behavioral difference: #482's `encode_address_header` also truncated bare emails at the first non-email character as post-CRLF-stripping cleanup; we rely on mail-builder's angle-bracket wrapping and Gmail's API validation to reject malformed addresses instead of silently truncating. See `test_mailbox_parse_strips_*` and `test_send_crlf_injection_*`.

### Note on #395

The open attachments PR hand-rolls multipart MIME on the old MessageBuilder. This refactor provides a cleaner foundation for that feature using mail-builder's native multipart support.

## Checklist:

### DIFF
--- a/.changeset/send-from-flag.md
+++ b/.changeset/send-from-flag.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": minor
+---
+
+Add `--from` flag to `+send` for send-as alias support
--- a/src/helpers/gmail/mod.rs
+++ b/src/helpers/gmail/mod.rs
@@ -810,6 +810,12 @@ impl Helper for GmailHelper {
                             .help("Email body (plain text, or HTML with --html)")
                             .required(true)
                             .value_name("TEXT"),
+                    )
+                    .arg(
+                        Arg::new("from")
+                            .long("from")
+                            .help("Sender address (for send-as/alias; omit to use account default)")
+                            .value_name("EMAIL"),
                     ),
             )
             .after_help(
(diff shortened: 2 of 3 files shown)


malob:
Superseded by #526

PR #494, opened 2026-03-15
fix(token_storage): propagate errors when creating/securing token dir

### LINKED ISSUE #493: Bug: Silent failures when creating token storage directory
I noticed in `src/token_storage.rs` that failures to create the token directory or set its permissions are silently ignored using `let _`.\n\n```rust\nlet _ = tokio::fs::create_dir_all(parent).await;\n// ...\nlet _ = std::fs::set_permissions(parent, std::fs::Permissions::from_mode(0o700));\n```\n\nIf directory creation fails (e.g. due to permissions), the subsequent file write will likely fail with a confusing error, or if permission setting fails, the token file might be left with insecure permissions.\n\nThese errors should probably be propagated up to the caller.

mango766:
Fixes #493\n\nPreviously, failures to create the token directory or set its permissions were silently ignored with `let _`. This could lead to confusing errors later when writing the token file, or worse, leave the directory with insecure permissions if the `chmod` fails. \n\nThis PR changes `let _` to `?` to ensure these errors are propagated.

### DIFF
--- a/src/token_storage.rs
+++ b/src/token_storage.rs
@@ -74,11 +74,11 @@ impl EncryptedTokenStorage {
         let encrypted = crate::credential_store::encrypt(json.as_bytes())?;
 
         if let Some(parent) = self.file_path.parent() {
-            let _ = tokio::fs::create_dir_all(parent).await;
+            tokio::fs::create_dir_all(parent).await?;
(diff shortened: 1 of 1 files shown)


jpoehnelt:
closing since cla is not signed

PR #504, opened 2026-03-15
fix(auth): robustly enforce readonly scopes by injecting prompt=consent

### LINKED ISSUE #168: gws auth login --readonly + auth export --unmasked appears to allow full access on external machine

dumko2001:
## Description

Robustly enforce readonly scopes by injecting `prompt=consent` into the OAuth flow when the `--readonly` flag is used. This ensures that Google always returns a refresh token and honors the requested readonly scopes, even if the user had previously authenticated with broader scopes.

The implementation:
- Updates `CliFlowDelegate` to handle a `force_consent` flag.
- Dynamically injects or modifies the `prompt` parameter in the OAuth URL to include `consent`, while preserving existing prompt values (e.g. `select_account`) using space-delimiting as per RFC 6749.
- Passes the `force_consent` flag to the authenticator based on the presence of `--readonly` in the login arguments.

Fixes #168

**Dry Run Output:**
N/A (Authentication flow change)

## Checklist:

### DIFF
--- a/src/auth_commands.rs
+++ b/src/auth_commands.rs
@@ -213,8 +213,12 @@ impl yup_oauth2::authenticator_delegate::InstalledFlowDelegate for CliFlowDelega
                         .map(|i| pos + i)
                         .unwrap_or(display_url.len());
                     let current_prompt = &display_url[pos + 7..end_pos];
-                    if !current_prompt.contains("consent") {
-                        let new_prompt = format!("{}%20consent", current_prompt);
+                    if !current_prompt.split("%20").any(|p| p == "consent") {
+                        let new_prompt = if current_prompt.is_empty() {
+                            "consent".to_string()
+                        } else {
+                            format!("{}%20consent", current_prompt)
+                        };
                         display_url.replace_range(pos + 7..end_pos, &new_prompt);
                     }
                 } else if display_url.contains('?') {


dumko2001:
Applied the suggested fix for robust `prompt=consent` injection. Now splitting by `%20` and performing an exact match check, while also handling empty prompt parameters.

jpoehnelt:
Superseded by #520 which takes a newer approach to readonly scope enforcement.

PR #517, opened 2026-03-16
feat(gmail): add --attachment flag to +send helper

### LINKED ISSUE #498: Feature: Add --attachment flag to gmail +send helper
## Problem

The `gmail +send` helper currently supports `--to`, `--subject`, `--body`, `--cc`, `--bcc`, and `--html` flags, but has **no support for file attachments**. The help text suggests using the raw API instead:

> TIPS: For attachments, use the raw API instead: `gws gmail users messages send --json '...'`

However, this workaround has a critical limitation: **the entire MIME message (including base64-encoded file contents) must be passed as a command-line argument**. On Windows, this hits the ~32,000 character limit (`ERROR: Argument list too long`), making it impossible to send even a single small PDF attachment via the raw API.

This also affects Linux/macOS systems with the default `ARG_MAX` limit when sending larger files.

## Real-world use case

I needed to send emails with PDF attachments (legal filings to a court) using `gws` as part of a CLI automation pipeline. The workflow:

1. Generate PDF documents programmatically
2. Send them as email attachments to specific recipients

The `+send` helper was perfect for the text part, but I had to fall back to Python's `google-api-python-client` to handle attachments — defeating the purpose of having a unified CLI tool.

## Proposed solution

Add a `--attachment <PATH>` flag to `gws gmail +send` that can be specified multiple times:

```bash
# Single attachment
gws gmail +send --to recipient@example.com --subject "Report" --body "See attached." --attachment ./report.pdf

# Multiple attachments
gws gmail +send --to recipient@example.com --subject "Documents" --body "See attached." \
  --attachment ./doc1.pdf \
  --attachment ./doc2.pdf
```

### Implementation notes

- Read the file(s) from disk and build the MIME multipart message internally (in Rust), avoiding command-line length limits entirely
- Auto-detect MIME type from file extension (or default to `application/octet-stream`)
- Use the filename as the attachment name by default, with an optional `--attachment-name` override
- The Gmail API accepts the full MIME message as base64url-encoded `raw` field — the current `+send` already constructs this for text-only messages, so this extends that logic

### Alternative considered

Supporting `@file` syntax for `--json` (like curl's `@filename`) would also solve the command-line length issue for the raw API approach. But a dedicated `--attachment` flag is more user-friendly and consistent with the helper's design philosophy of abstracting away RFC 2822 formatting.

## Environment

- gws version: 0.16.0
- OS: Windows 11 (where the issue is most severe due to lower argument length limits)
- Also reproducible on Linux/macOS with larger attachments

alimpolat:
## Summary

Adds `--attachment <PATH>` flag to the `gmail +send` helper, enabling email attachments without requiring manual RFC 2822/MIME construction or hitting command-line length limits.

- Multiple files supported via repeated `--attachment` flags
- MIME types auto-detected from file extension (PDF, DOCX, PNG, etc.)
- Builds `multipart/mixed` MIME message internally (no CLI argument size issues)
- Path validation: rejects traversal (`../`), control characters, directories, and missing files
- Falls back to existing simple message format when no attachments are provided

### Usage

```bash
# Single attachment
gws gmail +send --to alice@example.com --subject 'Report' --body 'See attached.' --attachment ./report.pdf

# Multiple attachments
gws gmail +send --to alice@example.com --subject 'Docs' --body 'Files attached.' \
  --attachment ./doc1.pdf \
  --attachment ./doc2.pdf
```

Closes #498

## Files Changed

| File | Change |
|------|--------|
| `src/helpers/gmail/mod.rs` | `MessageBuilder::build_with_attachments()`, `Attachment` struct, `mime_type_from_extension()`, `--attachment` CLI arg, tests |
| `src/helpers/gmail/send.rs` | `SendConfig.attachments`, `validate_attachment_path()`, `load_attachments()`, tests |
| `.changeset/add-attachment-flag.md` | Minor version bump changeset |

## Test plan

- [x] Existing `send.rs` tests still pass (parse args, HTML flag, cc/bcc)
- [x] New tests: attachment parsing, path traversal rejection, missing file rejection, directory rejection, control char rejection
- [x] New tests: `load_attachments()` reads file data and detects MIME type
- [x] New tests: `build_with_attachments()` produces valid multipart/mixed MIME
- [x] New tests: empty attachments falls back to simple message
- [x] New tests: `mime_type_from_extension()` common types, case insensitive, unknown fallback
- [x] New tests: `generate_mime_boundary()` uniqueness
- [ ] CI: `cargo test`, `cargo clippy`, codecov/patch

### DIFF
--- a/src/helpers/gmail/mod.rs
+++ b/src/helpers/gmail/mod.rs
@@ -627,21 +627,15 @@ impl MessageBuilder<'_> {
         // Attachment parts
         for att in attachments {
             let encoded = base64::engine::general_purpose::STANDARD.encode(&att.data);
-            // Escape backslashes and quotes per RFC 2183 to prevent
-            // malformed Content-Disposition headers from filenames like:
-            //   my"file.txt  →  my\"file.txt
-            let safe_filename = sanitize_header_value(&att.filename)
-                .replace('\\', "\\\\")
-                .replace('"', "\\\"");
+            let disposition = encode_content_disposition(&att.filename);
             message.push_str(&format!(
                 "--{boundary}\r\n\
                  Content-Type: {mime_type}\r\n\
                  Content-Transfer-Encoding: base64\r\n\
-                 Content-Disposition: attachment; filename=\"{filename}\"\r\n\
+                 {disposition}\r\n\
                  \r\n\
                  {encoded}\r\n",
                 mime_type = att.mime_type,
-                filename = safe_filename,
             ));
         }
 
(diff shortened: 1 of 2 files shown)


alimpolat:
Hi! All the Gemini review feedback has been addressed across the follow-up commits:

- **TOCTOU** — Moved path validation into `load_attachments()`, immediately before `fs::read()`, to minimize the race window.
- **Non-ASCII filenames** — Added RFC 2231 encoding (`filename*=UTF-8''...`) with ASCII fallback for older clients.
- **DRY headers** — Extracted `build_common_headers()` shared by `build()` and `build_with_attachments()`.
- **Filename escaping** — Backslash and quote escaping per RFC 2183.
- **Symlink safety** — `canonicalize()` + CWD containment check; absolute paths rejected.
- **Duplicate builder** — Single `MessageBuilder` instance before the conditional.

All 160 Gmail-related tests pass locally (including 25+ new tests for attachments, path validation, MIME encoding, and non-ASCII filenames).

The `Format` check is skipping — I believe it needs a maintainer to approve the workflow for first-time contributors. Would appreciate a review when you get a chance. Thanks!

malob:
FYI, this was superseded by #526 which added attachment support built on top of the mail-builder crate.

PR #537, opened 2026-03-18
fix(client): add 10s connect timeout to prevent initial hangs

dumko2001:
This PR adds a 10-second connection timeout to the HTTP client to prevent the CLI from hanging indefinitely on initial handshakes (DNS/TCP/TLS). 

Note: A global request timeout was intentionally omitted to ensure that large file operations (e.g., Drive uploads/downloads) and long-running API streams (e.g., Gmail watch) are not prematurely terminated.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -5,7 +5,6 @@ const MAX_RETRIES: u32 = 3;
 /// or misconfigured server from hanging the process indefinitely.
 const MAX_RETRY_DELAY_SECS: u64 = 60;
 const CONNECT_TIMEOUT_SECS: u64 = 10;
-const REQUEST_TIMEOUT_SECS: u64 = 600;
 
 pub fn build_client() -> Result<reqwest::Client, crate::error::GwsError> {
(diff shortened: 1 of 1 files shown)


dumko2001:
Updated to use named constants (`CONNECT_TIMEOUT_SECS`, `REQUEST_TIMEOUT_SECS`) for the timeout values as suggested.

dumko2001:
Removed the global request timeout as suggested. This avoids breaking long-running operations like large file downloads while still providing protection against connection hangs via the 10s `connect_timeout`.

dumko2001:
Re-added a 10-minute request timeout to prevent infinite hangs while still accommodating large file transfers, as suggested.

dumko2001:
I kept the 10s connection timeout to
  prevent initial hangs (DNS/TCP/TLS), but removed the global request timeout. This
  provides the necessary safety against unresponsive servers without breaking support
  for large file operations or long-running streams that might legitimately exceed a
  10-minute limit.

PR #545, opened 2026-03-18
fix(auth): move keyring backend status message to debug log

dumko2001:
## Description

Previously, the message indicating which keyring backend was being used was printed to `stderr` on every invocation. This pollutes the output in automated environments and agent pipelines. This PR moves that message to a debug-level log using the `tracing` crate.

## Checklist:

### DIFF
--- a/.changeset/fix-keyring-logging.md
+++ b/.changeset/fix-keyring-logging.md
@@ -1,1 +1,1 @@
----\n"gws": patch\n---\n\nfix(auth): move keyring backend status message to debug log to reduce stderr pollution
+---\n"@googleworkspace/cli": patch\n---\n\nfix(auth): move keyring backend status message to debug log to reduce stderr pollution


madisonrickert:
Hey, just wanted to add a +1 for this fix. The `Using keyring backend: keyring` message on stderr causes issues when piping `gws` output through JSON parsers in automation pipelines. For example:

```bash
gws docs documents get --params '{"documentId": "..."}' --format json | python3 -c "import json, sys; doc = json.load(sys.stdin)"
```

If stderr accidentally gets merged (e.g. `2>&1`), the non-JSON line breaks the parser. Even when stderr is kept separate, it creates noise in CI logs and agent tool output.

Moving this to debug-level logging seems like the right approach — it's useful for troubleshooting but shouldn't appear during normal operation. Would love to see this reopened or the change landed in a future release.

PR #564, opened 2026-03-19
fix(auth): stop TUI from unconditionally injecting cloud-platform scope

### LINKED ISSUE #562: bug: Interactive TUI unconditionally injects cloud-platform scope, breaking org-restricted accounts
## Problem

The interactive TUI scope picker in `gws auth login` unconditionally injects the `cloud-platform` scope after selection, regardless of what the user chose. This causes `admin_policy_enforced` errors for users in Google Workspace organizations that block the `cloud-platform` scope.

## Root Cause

In `src/auth_commands.rs`, line ~860:

```rust
// Always include cloud-platform scope
if !selected.contains(&PLATFORM_SCOPE.to_string()) {
    selected.push(PLATFORM_SCOPE.to_string());
}
```

This runs after every TUI selection, forcing `cloud-platform` into the scope list even when the user explicitly chose only basic services.

## Contradiction with DEFAULT_SCOPES

The code comments on lines 56-62 explicitly acknowledge that `cloud-platform` is restricted and was intentionally removed from `DEFAULT_SCOPES`:

```rust
/// Previously this included `pubsub` and `cloud-platform`, which Google marks
/// as *restricted* and blocks for unverified apps, causing `Error 403:
/// restricted_client`.  Use `--scopes` to add those scopes explicitly when you
/// have a verified app or a GCP project with the APIs enabled and approved.
pub const DEFAULT_SCOPES: &[&str] = MINIMAL_SCOPES;
```

Yet the TUI path contradicts this design by force-injecting the scope that was deliberately excluded.

## Reproduction

```bash
# This works (no TUI, no cloud-platform injected):
echo "" | gws auth login

# This fails (TUI injects cloud-platform):
gws auth login

# This also fails (TUI still opens even with -s):
gws auth login -s drive,gmail,calendar,sheets,docs,slides
```

### Error

```
Error 400: admin_policy_enforced
```

The generated OAuth URL contains `cloud-platform` even when only basic services (drive, gmail, calendar) are selected in the TUI.

## Expected Behavior

- The TUI should NOT force-inject `cloud-platform` — it should respect the user's selection
- `cloud-platform` should only be included when the user explicitly selects it in the TUI, uses `--full`, or adds it via `--scopes`
- The TUI behavior should be consistent with `DEFAULT_SCOPES` (which excludes `cloud-platform`)

## Suggested Fix

Remove or conditionalize the forced injection at line ~860:

```diff
-            // Always include cloud-platform scope
-            if !selected.contains(&PLATFORM_SCOPE.to_string()) {
-                selected.push(PLATFORM_SCOPE.to_string());
-            }
```

If `cloud-platform` needs to remain available, it should be a selectable option in the TUI rather than being silently injected.

## Workaround

Bypass the TUI by piping empty input:

```bash
echo "" | gws auth login
```

## Environment

- gws v0.18.0 (homebrew)
- Managed Google Workspace account
- macOS

anshul-garg27:
## Summary

The interactive TUI scope picker unconditionally injects `cloud-platform` after the user's selection (line ~835 in `auth_commands.rs`), even when the user only picked basic services like Drive or Gmail. This contradicts `DEFAULT_SCOPES` which deliberately excludes `cloud-platform` to avoid `restricted_client` / `admin_policy_enforced` errors on managed Workspace accounts.

**Changes:**
- Remove the forced `cloud-platform` injection after TUI selection
- Remove the `cloud-platform` auto-pass-through in `scope_matches_service` so it's filtered like any other scope when `--services` is used
- Clean up the now-unused `PLATFORM_SCOPE` constant

`cloud-platform` remains available in the TUI picker as a selectable option and is still included when using `--full`. Users who need it can select it explicitly.

Closes #562

## Test plan

- [x] `cargo test` — 749 pass, 0 fail
- [x] Existing `scope_matches_service` test updated to verify cloud-platform is NOT auto-included
- [x] Services filter integration tests updated to reflect the fix

### DIFF
--- a/.changeset/fix-tui-cloud-platform-scope.md
+++ b/.changeset/fix-tui-cloud-platform-scope.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Stop unconditionally injecting cloud-platform scope in TUI picker
--- a/src/auth_commands.rs
+++ b/src/auth_commands.rs
@@ -539,18 +539,10 @@ async fn resolve_scopes(
 /// `https://www.googleapis.com/auth/`). A scope matches a service if its
 /// short name equals the service or starts with `service.` (e.g. service
 /// `drive` matches `drive`, `drive.readonly`, `drive.metadata.readonly`).
-///
-/// The `cloud-platform` scope always passes through since it's a
-/// cross-service platform scope.
 fn scope_matches_service(scope_url: &str, services: &HashSet<String>) -> bool {
     let short = scope_url
         .strip_prefix("https://www.googleapis.com/auth/")
         .unwrap_or(scope_url);
-
-    // cloud-platform is a cross-service scope, always include
-    if short == "cloud-platform" {
-        return true;
-    }
 
     let prefix = short.split('.').next().unwrap_or(short);
 
@@ -660,7 +652,7 @@ fn run_discovery_scope_picker(
     relevant_scopes: &[crate::setup::DiscoveredScope],
     services_filter: Option<&HashSet<String>>,
 ) -> Option<Vec<String>> {
-    use crate::setup::{ScopeClassification, PLATFORM_SCOPE};
+    use crate::setup::ScopeClassification;
     use crate::setup_tui::{PickerResult, SelectItem};
 
     let mut recommended_scopes = vec![];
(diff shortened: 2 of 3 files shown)


VictorML11:
Hi, I noticed this PR was closed. Is there any plan to reopen it or continue this work? I'm interested in seeing this land and would be happy to help test or provide feedback if that would help move things forward. Thanks\!

anshul-garg27:
hey, reopened as #662 — rebased on latest main since the repo got restructured into a workspace layout. same fix, clean CI.

PR #568, opened 2026-03-19
fix(gmail): add HTML draft creation guidance to generated skill

kleinq:
## Problem

When creating Gmail drafts via `gws gmail users drafts create`, users naturally
pass plain-text bodies in the raw RFC 2822 message. Plain-text emails are
line-wrapped by email clients (typically at 72–76 chars), breaking paragraph
flow and making the draft look unprofessional.

## Fix

Add a `## Creating Drafts` section to the generated `gws-gmail` skill that:

- Explains the wrapping problem and why HTML is required
- Shows the exact paragraph/newline → `<div>`/`<br>` conversion pattern
  (matching how Gmail's native compose formats messages)
- Provides a copy-paste Python snippet to build and base64url-encode the
  RFC 2822 message with `Content-Type: text/html`

This is a gmail-specific section (guarded by `alias == "gmail"`) so it only
appears in the `gws-gmail` skill, not in other service skills.

## Testing

Run `gws generate-skills` after this change; the generated
`skills/gws-gmail/SKILL.md` should contain the new `## Creating Drafts`
section before `## Discovering Commands`.

### DIFF
--- a/src/generate_skills.rs
+++ b/src/generate_skills.rs
@@ -474,6 +474,40 @@ metadata:
         }
     }
 
+    // Gmail-specific: explain how to create properly formatted drafts
+    if alias == "gmail" {
+        out.push_str("## Creating Drafts\n\n");
+        out.push_str("**Always use HTML when creating drafts** to prevent email clients from wrapping long lines. Build the raw RFC 2822 message with `Content-Type: text/html` and convert the plain text body to HTML:\n\n");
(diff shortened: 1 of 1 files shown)


malob:
FWIW, I've got adding draft support for the `+send`, `+reply`, `+reply-all`, and `+forward` helpers (which all support sending sending HTML emails using the `--html` flag) ready to go: https://github.com/malob/cli/commit/93997d93a1c5ecfed9235004b60c4da1f80fdb0a

(Just waiting on another PR to land which touches at lot of the same code.)

Issue #555, opened 2026-03-18
Feature request: +draft helper and CC/threading support for +send


fragileshark:
## Summary

The `+send` helper is great for simple emails, but there's no equivalent for creating drafts, and `+send` lacks support for CC/BCC, `In-Reply-To`, `References`, and thread IDs.

## Current behavior

- `gws gmail +send` only supports `--to`, `--subject`, `--body`
- No `+draft` helper exists
- Creating drafts requires manually constructing RFC 2822 messages, base64-encoding them, and calling `gws gmail users drafts create --json '...'` — which is error-prone (encoding issues cause drafts to appear in Gmail with invisible bodies)

### 1. Add `+draft` helper

Mirror `+send` but target `drafts/create` instead of `messages/send`:

```bash
gws gmail +draft --to alice@example.com --subject "Hello" --body "Hi Alice!"
```

### 2. Add shared flags to both `+send` and `+draft`

| Flag | Purpose |
|------|---------|
| `--cc EMAIL` | CC recipient |
| `--bcc EMAIL` | BCC recipient |
| `--in-reply-to ID` | In-Reply-To header (for threading) |
| `--references ID` | References header (for threading) |
| `--thread-id ID` | Gmail thread ID (for replies) |

### Example usage

```bash
# Simple draft
gws gmail +draft --to alice@example.com --subject "Hello" --body "Hi Alice!"

# Reply draft in a thread with CC
gws gmail +draft \
  --to alice@example.com \
  --cc bob@example.com \
  --subject "Re: Project update" \
  --body "Sounds good, let's proceed." \
  --thread-id 19d01b52a042ad59 \
  --in-reply-to "<message-id@mail.gmail.com>"
```

## Motivation

I use the gws CLI heavily via LLM agents (Claude Code) for email workflows. The agents need to create drafts for human review before sending. Right now the only path is raw API calls with manual base64 encoding, which is brittle — the encoding has to be done in a very specific way or Gmail's UI renders the draft with an empty body. A `+draft` helper that handles the RFC 2822 formatting and encoding (like `+send` already does) would make this reliable.

The CC/threading flags would also benefit `+send` directly — right now any reply or CC'd email requires dropping down to the raw API.

malob:
Just a heads up — most of what you're describing already exists in recent releases:

[...]

Issue #508, opened 2026-03-15
Feature: Possibly more gfranular approach for email labeling/drafting?


keramblock:
One of the areas where your product could really shine is email classification and triage — agents, automations, and tools that read your inbox and organize it with labels. But currently there's no way to do that safely without granting automation rights to send/delete emails.

What would help
  A new scope (e.g. gmail.labels.apply or gmail.organize) that permits:
  - users.messages.modify — limited to addLabelIds / removeLabelIds
  - users.messages.batchModify — same constraint

  Without granting:
  - users.messages.send
  - users.messages.delete / users.messages.trash
  - users.drafts.send

Could I ask you to take a look on this? 
P.S. Same for the drafts, looks like I cant allow creating of the drafts without also allowing to send it
P.P.S. Thank you for the great program!

m-roberts:
Would love the ability to create/update/delete drafts without also granting send permissions.

PR #499, opened 2026-03-15
feat(gmail): implement robust draft-only mode with centralized policy enforcement (#424)

dumko2001:
## Description

Implemented a robust Gmail `draft-only` mode to prevent accidental email sends. This update addresses a critical security bypass where helper commands (like `+send`, `+reply`, `+forward`) previously circumvented the safety flag. 

**Key Changes:**
- **Centralized Enforcement**: Moved policy checks into `executor.rs` and Gmail's `send_raw_email` to ensure all interaction layers (raw API and helpers) are protected.
- **Architectural Consolidation**: Introduced a unified `ExecutionPolicy` struct to manage both Model Armor sanitization and Gmail safety settings consistently across all service helpers.
- **Early Rejection**: Enforced checks before authentication in `main.rs` to provide a better user experience for blocked operations.
- **Fixes**: Removed a duplicate disclaimer bug in `print_usage` and resolved several compiler warnings.

Fixes #424

**Dry Run Output:**
```json
{
  "body": {
    "raw": "VG86IGFsaWNlQGV4YW1wbGUuY29tDQpTdWJqZWN0OiBEcmFmdA0KTUlNRS1WZXJzaW9uOiAxLjANCkNvbnRlbnQtVHlwZTogdGV4dC9wbGFpbjsgY2hhcnNldD11dGYtOA0KDQpUaGlzIGlzIHNhZmU="
  },
  "dry_run": true,
  "is_multipart_upload": false,
  "method": "POST",
  "query_params": [],
  "url": "https://gmail.googleapis.com/gmail/v1/users/me/messages/send"
}
```

## Checklist:

- [x] My code follows the `AGENTS.md` guidelines (no generated `google-*` crates).
- [x] I have run `cargo fmt --all` to format the code perfectly.
- [x] I have run `cargo clippy -- -D warnings` and resolved all warnings.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have provided a Changeset file (e.g. via `pnpx changeset`) to document my changes.

Issue #424, opened 2026-03-12
Feature request: Gmail draft-only mode for CLI+agent workflows


escape0707:
Please consider adding a Gmail draft-only mode/policy to `gws`.

Desired behavior:
- allow reading Gmail as configured
- allow creating/updating drafts
- block all send paths

In particular, this mode should prevent:
- `users.messages.send`
- `users.drafts.send`
- reply / reply-all / forward actions that result in sending
- any equivalent helper/tool aliases

A common agent workflow is: let the agent prepare Gmail drafts for human review, but guarantee that it cannot actually send mail.

That “draft yes, send no” boundary would be very useful for agent integrations.

Notes:

I understand this may depend on Gmail/OAuth/API limitations. If true Google-enforced draft-only permission is not currently possible, a `gws`-side policy/tool filter would still be very valuable.

If this request belongs in another repo/team, please feel free to redirect it.

jpoehnelt:
Closing — PR #499 was closed. The simpler --draft flag approach in PR #571 covers this use case without the architectural scope creep. See [helper guidelines](https://github.com/googleworkspace/cli/blob/main/src/helpers/README.md).

escape0707:
@jpoehnelt 

Like what have been said at https://github.com/googleworkspace/cli/pull/571#issuecomment-4114717413

I don't think that PR addresses my problem. I want a config that can block gws from send mail but the PR that claims it fix my issue only provided an opt-in flag and agents can forget to it.

And demanding users to hand craft rules to block sending command is just an ugly hack for my taste. I think it's not a stable solution and put unnecessary maintenance burden to each user. Also a whitelist rules pattern suggested by the PR creator is quite inconvenient and inconsistent.

If you agree with me, I suggest to reopen this issue.

PR #571, opened 2026-03-20
feat(gmail): add --draft flag to +send, +reply, +reply-all, +forward

malob:
## Description

Adds a `--draft` flag to `+send`, `+reply`, `+reply-all`, and `+forward`. When set, calls `users.drafts.create` instead of `users.messages.send`. Message construction is identical; only the API method and metadata wrapper change.

### How it works

- `--draft` added to `common_mail_args` (one place, all four helpers get it)
- `resolve_draft_method` navigates `users.drafts.create` in the discovery doc
- `resolve_mail_method` dispatcher selects between send and draft based on the flag
- `dispatch_raw_email` (renamed from `send_raw_email` to reflect its dual role) switches the method and wraps draft metadata in the `{"message": {...}}` envelope required by the Drafts API
- Threaded drafts (replies and forwards) preserve `threadId` in the metadata
- After successful draft creation, a tip is printed to stderr with the `users.drafts.send` command (suppressed during `--dry-run`)

### Related issues and PRs

- **#424** / **PR #499**: Requests a draft-only *policy* that blocks all send paths for agent safety. The `--draft` flag handles the draft creation side, so the policy work can focus purely on enforcement, restricting which operations are allowed, without needing to implement how drafts are created.

- **#286**: The original "can I draft without sending?" request. The `--draft` flag answers the drafting half; the "without sending" enforcement is #424's territory.

- **#555** (closed): Requested a `+draft` helper and CC/threading support for `+send`. CC/BCC and threading already exist via `+reply`/`+reply-all`/`+forward`; this PR adds the draft capability. Reporter closed the issue after learning about the existing features.

- **PR #568**: Adds skill guidance for manually constructing RFC 2822 messages for `users.drafts.create`. The `--draft` flag makes that workflow significantly simpler: `+send --draft` handles all the RFC 5322 formatting, MIME encoding, and base64 automatically.

- **#508**: Requests finer-grained Gmail OAuth scopes for label-only and draft-without-send access. `gmail.labels` covers CRUD on label definitions, but applying labels to messages requires `gmail.modify`, which also grants send/compose. And `gmail.compose` bundles draft creation and sending together. So there is no scope combination that allows label application or draft creation without also granting send. Client-side enforcement like the draft-only policy from #424 is the available workaround.

### Live testing

Tested against a real Gmail account:

| # | Test | Result |
|---|------|--------|
| 1 | `+send --draft` | Draft created with `DRAFT` label |
| 2 | `+reply --draft` | Threaded draft in correct thread |
| 3 | `+send --draft --dry-run` | Dry-run output, no tip printed |
| 4 | `+send` (without `--draft`) | Sends normally (no regression) |

### Test coverage

697 total tests (6 new). New tests cover:
- `build_send_metadata` for all four `(thread_id, draft)` combinations
- `resolve_draft_method` discovery doc traversal
- Absence of `threadId` in draft metadata when no thread

**Dry Run Output:**
```json
{
  "body": "{\"message\":{}}",
  "dry_run": true,
  "is_multipart_upload": true,
  "method": "POST",
  "query_params": [],
  "url": "https://gmail.googleapis.com/upload/gmail/v1/users/me/drafts"
}
```

## Checklist:

### DIFF
--- a/.changeset/gmail-draft-flag.md
+++ b/.changeset/gmail-draft-flag.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": minor
+---
+
+Add `--draft` flag to Gmail `+send`, `+reply`, `+reply-all`, and `+forward` helpers to save messages as drafts instead of sending them immediately
--- a/crates/google-workspace-cli/src/helpers/gmail/forward.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/forward.rs
@@ -84,7 +84,7 @@ pub(super) async fn handle_forward(
 
     let raw = create_forward_raw_message(&envelope, &original, &all_attachments)?;
 
-    super::send_raw_email(
+    super::dispatch_raw_email(
         doc,
         matches,
         &raw,
@@ -510,7 +510,8 @@ mod tests {
                 Arg::new("no-original-attachments")
                     .long("no-original-attachments")
                     .action(ArgAction::SetTrue),
-            );
+            )
+            .arg(Arg::new("draft").long("draft").action(ArgAction::SetTrue));
         cmd.try_get_matches_from(args).unwrap()
     }
 
--- a/crates/google-workspace-cli/src/helpers/gmail/mod.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/mod.rs
@@ -42,6 +42,9 @@ use std::pin::Pin;
 
 pub struct GmailHelper;
 
+/// Broad scope used by reply/forward handlers for both message metadata
+/// fetching and the final send/draft operation. Covers `messages.send`,
+/// `drafts.create`, and read access in a single token.
 pub(super) const GMAIL_SCOPE: &str = "https://www.googleapis.com/auth/gmail.modify";
 pub(super) const GMAIL_READONLY_SCOPE: &str = "https://www.googleapis.com/auth/gmail.readonly";
 pub(super) const PUBSUB_SCOPE: &str = "https://www.googleapis.com/auth/pubsub";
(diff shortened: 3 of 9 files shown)


keramblock:
Hi @malob, thank you for taking on my issue as well as the others. 
I have a small question: 
Are you sure that "label management is already possible via the gmail.labels scope"? 

In my testing, gmail.labels only covers CRUD on label definitions (users.labels.*), not applying labels to messages — users.messages.modify and users.threads.modify both require gmail.modify, which also grants send/compose. 

This is exactly what I'm trying to avoid: I want an agent to organize my inbox with labels without it being able to create or send messages without my explicit permission.

P.S. that would be nice to force that as a draft only mode if possible to not rely on the llm not forgetting to add flag.

malob:
You're right. I misstated that. `gmail.labels` covers CRUD on label definitions (`users.labels.*`), but applying labels to messages via `users.messages.modify` requires `gmail.modify`. And `gmail.modify` also grants send/compose access, so there's no scope combination that allows both label application and draft creation without also granting send. I'll update the PR description to correct this.

On the draft-only mode point: agreed, `--draft` as a flag is opt-in, which means the agent can forget it. A few options for enforcement:

- The policy layer from #424 / PR #499 would block send paths at the `gws` level, regardless of what the agent does
- Many agent frameworks support command allowlists. For example, in Claude Code you could auto-approve only the draft variants:
  ```json
  {
    "permissions": {
      "allow": [
        "Bash(gws gmail +send --draft *)",
        "Bash(gws gmail +reply --draft *)",
        "Bash(gws gmail +reply-all --draft *)",
        "Bash(gws gmail +forward --draft *)"
      ]
    }
  }
  ```
  Non-draft commands would still require manual user approval. Not a hard block, but an additional safeguard.

jpoehnelt:
Merged #589. Some conflicts, might need to regenerate skills too

jpoehnelt:
⚠️ **Rebase required** — `src/` has moved to `crates/google-workspace-cli/src/` as part of a workspace refactor (#613). Please rebase on `main` to resolve path conflicts.

malob:
On it.

PR #595, opened 2026-03-23
feat(gmail): add --thread-id, --delivered-to, and --sent-last flags to +triage

eternalwaitt:
## Summary

`+triage` currently returns `id`, `from`, `subject`, and `date`. Three fields that are commonly needed for agent workflows and shared inbox routing are missing — even though they're already present in the single metadata fetch that `+triage` already performs.

This PR adds three opt-in flags:

| Flag | Field added | Source |
|---|---|---|
| `--thread-id` | `threadId` | Top-level field on every `messages.get` response |
| `--delivered-to` | `deliveredTo` | `Delivered-To` header (added to `metadataHeaders`) |
| `--sent-last` | `sentLast` (bool) | `true` when `SENT` is in `labelIds` |

**No extra API calls.** All three are derived from data already fetched in the existing per-message metadata request. `Delivered-To` is appended to the `metadataHeaders` list (one more string in the query param, same request).

All flags default to `false` — existing output is completely unchanged.

## Use cases

**`--thread-id`** — needed for any follow-up operation (reply, label, thread lookup) that requires a `threadId`. Without it, callers have to do a second fetch just to get the thread ID.

**`--delivered-to`** — shared inboxes and alias routing. When a single inbox receives mail for multiple aliases (e.g. `alice@`, `bob@`, `shared@`), the `To` header shows the original recipient while `Delivered-To` shows the actual delivery target. Grouping or routing by recipient requires this field.

**`--sent-last`** — detect threads where the current user is waiting for a reply, without a separate `threads.get` call. If `SENT` is in `labelIds`, the last action on that thread was an outbound message.

## Examples

```bash
# Include thread ID for downstream operations
gws gmail +triage --format json --thread-id | jq '.[].threadId'

# Group by delivery target in a shared inbox
gws gmail +triage --format json --delivered-to | jq 'group_by(.deliveredTo)'

# Surface threads waiting for a reply
gws gmail +triage --format json --sent-last | jq '[.messages[] | select(.sentLast)]'

# All three together
gws gmail +triage --format json --thread-id --delivered-to --sent-last
```

## Testing

Unit tests added for each new flag: default-off, set-when-passed, and all four opt-in flags combined.

### DIFF
--- a/src/helpers/gmail/mod.rs
+++ b/src/helpers/gmail/mod.rs
@@ -1240,6 +1240,24 @@ TIPS:
                         .help("Include label names in output")
                         .action(ArgAction::SetTrue),
                 )
+                .arg(
+                    Arg::new("thread-id")
+                        .long("thread-id")
+                        .help("Include threadId in output")
+                        .action(ArgAction::SetTrue),
+                )
+                .arg(
+                    Arg::new("delivered-to")
+                        .long("delivered-to")
+                        .help("Include Delivered-To header in output (useful for shared/alias inboxes)")
+                        .action(ArgAction::SetTrue),
+                )
+                .arg(
+                    Arg::new("sent-last")
+                        .long("sent-last")
+                        .help("Include sentLast boolean: true when the SENT label is present, indicating you sent the most recent message in the thread")
+                        .action(ArgAction::SetTrue),
+                )
                 .after_help(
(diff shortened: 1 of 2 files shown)


eternalwaitt:
I have signed the CLA. Please re-check.

PR #597, opened 2026-03-23
feat(gmail): add --thread-id, --delivered-to, and --sent-last flags to +triage

eternalwaitt:
## Summary

The `gmail +triage` helper fetches per-message metadata (From, Subject, Date) concurrently. Three pieces of data are already present in each API response but not exposed as output fields:

- `threadId` — the Gmail thread ID (already returned by the messages API)
- `Delivered-To` header — already fetched (added to `metadataHeaders` in this PR)
- `sentLast` — whether the `SENT` label is present on the message (already fetched via `labelIds`)

This PR adds three opt-in flags to expose them:

| Flag | Description |
|------|-------------|
| `--thread-id` | Include `threadId` field in each message entry |
| `--delivered-to` | Include `deliveredTo` field (from `Delivered-To` header) |
| `--sent-last` | Include `sentLast` boolean (true when message has `SENT` label) |

All flags are **off by default** — no change in output for existing callers. The `Delivered-To` header is now always fetched in `metadataHeaders` since it costs nothing extra and avoids a second request when `--delivered-to` is toggled.

## Use cases

- `--thread-id`: enables callers to group messages by thread or fetch thread details in a follow-up call
- `--delivered-to`: useful for shared inboxes where multiple aliases route to one mailbox (identifies which alias received the message)
- `--sent-last`: identifies threads where the authenticated user sent the last message — useful for "waiting for reply" classification without fetching full thread data

## Tests

Seven new unit tests cover: default-off behaviour for each flag, set-when-passed behaviour, and combined flag usage.

### DIFF
--- a/src/helpers/gmail/mod.rs
+++ b/src/helpers/gmail/mod.rs
@@ -1240,6 +1240,24 @@ TIPS:
                         .help("Include label names in output")
                         .action(ArgAction::SetTrue),
                 )
+                .arg(
+                    Arg::new("thread-id")
+                        .long("thread-id")
+                        .help("Include threadId in output")
+                        .action(ArgAction::SetTrue),
+                )
+                .arg(
+                    Arg::new("delivered-to")
+                        .long("delivered-to")
+                        .help("Include Delivered-To header in output (useful for shared/alias inboxes)")
+                        .action(ArgAction::SetTrue),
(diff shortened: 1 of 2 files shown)


jpoehnelt:
See src/helpers/README.md

PR #423, opened 2026-03-12
feat: add HTTP proxy support via environment variables

femto:
## Summary

When `http_proxy`/`https_proxy`/`all_proxy` environment variables are set, use `reqwest` (which natively supports proxy) for token refresh instead of `yup-oauth2`'s `hyper`-based client (which doesn't support proxy).

This enables `gws` to work in environments that require HTTP proxy to access Google APIs (e.g., users in China).

## Changes

- **Cargo.toml**: Enable `reqwest`'s default features including proxy and SOCKS support
- **src/auth.rs**: 
  - Add `has_proxy_env()` helper to detect proxy environment variables
  - Add `refresh_token_with_reqwest()` for proxy-aware token refresh
  - When proxy env vars are detected, use `reqwest` directly instead of waiting for `yup-oauth2`'s hyper client to timeout

## Testing

Tested in an environment requiring HTTP proxy:
```bash
export https_proxy=http://127.0.0.1:7890
gws drive files list --params '{"pageSize": 5}'
# Works! Completes in ~2 seconds
```

Without this change, the command would hang waiting for hyper's connection timeout, then fail.

Fixes #422

PR #622, opened 2026-03-25
feat: add HTTP proxy support via environment variables

femto:
## Summary

When `http_proxy`/`https_proxy`/`all_proxy` environment variables are set, use `reqwest` (which supports proxy env vars natively) for token refresh and the auth login token exchange instead of relying only on `yup-oauth2`'s `hyper`-based client.

This enables `gws` to work in environments that require an HTTP proxy to access Google APIs.

Supersedes closed PR #423 after rebasing onto the cargo workspace refactor.

## Changes

- `crates/google-workspace-cli/src/auth.rs`
  - Add shared proxy env detection
  - Add proxy-aware token refresh via `reqwest`
  - Improve error-body handling for failed token refresh responses
- `crates/google-workspace-cli/src/auth_commands.rs`
  - Add proxy-aware OAuth login flow for auth code exchange
  - Extract the non-proxy yup-oauth flow into a helper for readability
  - Reuse shared proxy/error helpers and keep the newer workspace auth scope behavior intact
- `crates/google-workspace-cli/Cargo.toml`
  - Keep `reqwest` configured with SOCKS support in the workspace crate
- `.changeset/proxy-support-review-fixes.md`
  - Include a patch changeset for the CLI package

## Testing

- `cargo clippy -p google-workspace-cli -- -D warnings`
- `cargo test -p google-workspace-cli auth::tests::`
- `cargo test -p google-workspace-cli auth_commands::tests::`

### DIFF
--- a/crates/google-workspace-cli/src/auth.rs
+++ b/crates/google-workspace-cli/src/auth.rs
@@ -51,7 +51,7 @@ async fn refresh_token_with_reqwest(
     client_secret: &str,
     refresh_token: &str,
 ) -> anyhow::Result<String> {
-    let client = reqwest::Client::new();
+    let client = crate::client::shared_client().map_err(anyhow::Error::from)?;
     let params = [
         ("client_id", client_id),
         ("client_secret", client_secret),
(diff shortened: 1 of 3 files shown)


femto:
Addressed the reqwest client reuse feedback by adding a shared client helper backed by std::sync::OnceLock in crates/google-workspace/src/client.rs and switching the proxy auth flows to use it.

PR #678, opened 2026-04-06
fix(helpers): respect --format flag in sheets +append, +read, and docs +write

nuthalapativarun:
## Description

`sheets +append`, `sheets +read`, and `docs +write` all hard-coded `OutputFormat::default()` (JSON) when calling `executor::execute_method`, meaning `--format table`, `--format yaml`, and `--format csv` had no effect on their output.

Each handler now reads the global `--format` flag from `ArgMatches` and passes it through to the executor, matching the existing behaviour of `calendar +agenda` and `gmail +triage`.

**Example before fix:**
```bash
# --format table was silently ignored; output was always JSON
gws sheets +read --spreadsheet ID --range Sheet1 --format table
```

**After fix:** output respects the `--format` flag as with all other commands.

**Dry Run Output:**
```json
// Not applicable — this is a bug fix with no API call changes.
// The --format flag controls output formatting, not request construction.
```

## Checklist:

- [ ] I have run `cargo fmt --all` to format the code perfectly.
- [ ] I have run `cargo clippy -- -D warnings` and resolved all warnings.

### DIFF
--- a/.changeset/helpers-respect-format-flag.md
+++ b/.changeset/helpers-respect-format-flag.md
@@ -1,0 +1,12 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Fix `--format` flag being silently ignored in sheets and docs helpers
+
+`sheets +append`, `sheets +read`, and `docs +write` all hard-coded
+`OutputFormat::default()` (JSON) when calling `executor::execute_method`,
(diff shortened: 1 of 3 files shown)


nuthalapativarun:
bot: rescan

PR #683, opened 2026-04-07
fix(gmail): preserve HTML body when text/html part has Content-ID (Outlook)

malob:
## Description

Fixes a bug where `+reply`, `+reply-all`, and `+forward` silently drop the HTML body from Outlook/Exchange messages, falling back to a plain-text conversion that loses all formatting, nested blockquotes, and inline images.

### The problem

Outlook/Exchange wraps messages in `multipart/related` so the HTML body can reference signature images via `cid:` URLs. As part of this structure, Outlook adds a `Content-ID` header to the `text/html` body part itself:

```
multipart/related
├── multipart/alternative
│   ├── text/plain     (no Content-ID)
│   └── text/html      (Content-ID: <ABC@exchange.example.com>)  ← dropped
├── image/png          (Content-ID, attachmentId)  ← signature images
└── ...
```

The MIME payload walker's `is_body_text_part` classification required `content_id_header.is_none()`, so Outlook HTML bodies were silently skipped. `body_html` ended up `None`, and `resolve_html_body` fell back to converting the plain text with `<br>` tags.

### How we found it

Replied to an Outlook/Exchange email using `+reply`. The quoted content in the sent message was a flat wall of `<br>`-separated text instead of the original's nested HTML blockquotes with styling. Inspected the original message via the Gmail API and found the `Content-ID` header on the `text/html` part.

### The fix

Replace the `content_id_header.is_none()` condition with an explicit MIME type allowlist (`text/plain` or `text/html`). Body parts are now identified by mime type + inline `body.data` + no `attachmentId` + no filename, regardless of Content-ID.

This is a tighter predicate than the old code: previously, any `text/*` subtype without Content-ID (e.g. `text/calendar`) would have passed the outer `is_body_text_part` check, though the inner branch already only handled `text/plain` and `text/html`. The new check makes the outer condition honest about what the inner branch requires.

### Additional improvement: warning for dropped inline-data leaves

Non-text, non-hydratable leaf parts with inline `body.data` (e.g. an `application/json` part without an `attachmentId`) were previously dropped with no diagnostic. A warning is now logged to stderr with the MIME type, filename, and size, making silent data loss debuggable.

### Verified with real Outlook email

Created forward drafts of the same Outlook message using the installed (broken) and fixed builds, then compared the raw HTML:

| | Blockquotes | HTML size | Result |
|---|---|---|---|
| **Before fix** | 0 | 5.9 KB | Flat `<br>` text, all nesting lost |
| **After fix** | 6 | 20 KB | Full HTML with nested blockquotes and styling |

### Test coverage

704 total tests (3 new). New tests:

- **`test_extract_payload_contents_plain_text_with_content_id`** — `text/plain` part with Content-ID is still recognized as body text (documents the behavioral change for both MIME types, not just `text/html`)
- **`test_parse_original_message_html_with_content_id_end_to_end`** — End-to-end regression test through `parse_original_message` → `resolve_html_body` with a realistic Outlook MIME structure. Verifies HTML body is preserved, `resolve_html_body` returns actual HTML (not `<br>`-converted fallback), and inline signature images are still collected as parts.
- **`test_extract_payload_contents_multiple_html_leaves_first_wins`** — Documents the current DFS first-wins heuristic when multiple `text/html` parts are eligible, without claiming it is semantically ideal (full `multipart/related` `start=` parameter support would be needed for that).

### Known limitations (pre-existing, not introduced by this fix)

- **DFS first-wins ordering for multiple eligible HTML leaves.** The walker takes the first `text/html` part it encounters in depth-first order rather than honoring `multipart/related`'s `start=` parameter for root-part resolution. In practice this is correct for all standard email client MIME structures — the first `text/html` in `multipart/alternative` is always the intended body. The `start=` parameter is rarely used in email (it's more common in SOAP/MTOM). Documented and tested as a heuristic, not a specification guarantee.

- **`decode_text_body` assumes UTF-8.** Non-UTF-8 text parts (ISO-8859-1, Windows-1252, Shift_JIS) are warned and dropped, falling back to plain text or snippet. The Gmail API does not normalize charset — the raw encoding from the sending MTA is preserved. A proper fix would read the `charset` parameter from the `Content-Type` header and transcode via a crate like `encoding_rs`. This primarily affects messages from older mail systems; most modern clients send UTF-8.

- **Non-text inline-data leaves are warned and dropped.** Parts like `text/calendar` or `application/json` with inline `body.data` but no `attachmentId` are logged to stderr but not preserved as attachments. The `OriginalPart` pipeline assumes `attachmentId` for later binary hydration, so treating inline-data parts as attachments would require a different fetch path. In practice, the Gmail API almost always externalizes binary data behind `attachmentId` — we tested with a 26-byte file via `messages.insert` and it still got an `attachmentId`.

## Checklist:

### DIFF
--- a/.changeset/fix-html-body-content-id.md
+++ b/.changeset/fix-html-body-content-id.md
@@ -1,0 +1,11 @@
+---
+"@googleworkspace/cli": patch
+---
+
+fix(gmail): preserve HTML body when text/html part has Content-ID
+
+Outlook/Exchange adds a Content-ID header to the text/html body part for
+multipart/related referencing. The MIME walker incorrectly treated any part
+with Content-ID as a non-body part, causing the HTML body to be silently
+dropped. Replies to Outlook messages fell back to a plain-text conversion,
+losing all formatting, nested blockquotes, and inline images.
--- a/crates/google-workspace-cli/src/helpers/gmail/mod.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/mod.rs
@@ -919,9 +919,15 @@ fn extract_payload_recursive(
     // Primary signal: does this part have fetchable binary data?
     let is_hydratable = !attachment_id.is_empty();
 
-    // A body text part has inline body.data, no attachmentId, no filename, and no Content-ID.
+    // A body text part is text/plain or text/html with inline body.data, no
+    // attachmentId, and no filename.
+    // Note: Content-ID is NOT checked here. Outlook/Exchange adds Content-ID to
+    // the text/html body part for multipart/related referencing — excluding parts
+    // with Content-ID would cause the HTML body to be silently dropped, falling
+    // back to a plain-text conversion that loses all formatting and nested quotes.
+    let is_text_mime = mime_type == "text/plain" || mime_type == "text/html";
     let is_body_text_part =
-        !is_hydratable && filename.is_empty() && content_id_header.is_none() && body_data.is_some();
+        !is_hydratable && filename.is_empty() && is_text_mime && body_data.is_some();
 
     if is_body_text_part {
         // body_data is guaranteed Some by the is_body_text_part check above.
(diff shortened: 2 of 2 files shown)


malob:
Bump to keep open — awaiting maintainer review.

PR #684, opened 2026-04-07
fix(gmail): use OAuth userinfo endpoint for From display name fallback

### LINKED ISSUE #644: `gmail +send` shows "grant profile scope" tip and sends with null From name even when profile scope is granted
## Summary

`gws gmail +send` always prints the tip `Tip: run gws auth login and grant the "profile" scope to include your display name in the From header`, and sends emails with a null display name in the From field — even when the `userinfo.profile` scope is already granted.

## Steps to Reproduce

1. Authenticate with profile scope:
   ```
   gws auth login
   ```
2. Confirm profile scope is present in the auth response:
   ```json
   {
     "scopes": [
       "https://www.googleapis.com/auth/gmail.modify",
       "openid",
       "https://www.googleapis.com/auth/userinfo.email",
       "https://www.googleapis.com/auth/userinfo.profile"
     ]
   }
   ```
3. Send an email:
   ```
   gws gmail +send --to "yourself@example.com" --subject "test" --body "Testing from header"
   ```
4. Observe the tip is still printed to stderr.
5. Read the sent message:
   ```
   gws gmail +read --id <message_id> --format json
   ```
6. Observe `from.name` is `null`:
   ```json
   {
     "from": {
       "name": null,
       "email": "user@example.com"
     }
   }
   ```

## Expected Behavior

- The tip should not appear when the profile scope is already granted.
- The From header should include the user's display name (fetched via the People/UserInfo API using the granted profile scope).

## Actual Behavior

- The tip prints unconditionally on every `+send` invocation.
- The From name is always null regardless of granted scopes.

## Environment

- **gws version:** 0.22.3
- **OS:** macOS 15.7.5 (arm64 / Apple Silicon)
- **Node:** v20.18.2
- **Keyring backend:** keyring (OS keychain)
- **Account type:** Google Workspace (enterprise)

malob:
## Description

Fixes the From display name fallback for Workspace accounts where the sendAs identity has no display name configured. The previous fallback called the People API (`people.googleapis.com/v1/people/me`), which requires the People API to be separately enabled in the GCP project. When it isn't enabled, the 403 `accessNotConfigured` error was misreported as "grant the profile scope" — even when the `userinfo.profile` scope was already present in the token.

### Root cause

The People API has two prerequisites: the OAuth scope (`userinfo.profile`) AND the API enabled in the GCP project console. The error handler at the old `fetch_profile_display_name` only checked for 403 and assumed it meant "missing scope," which is wrong when the scope is granted but the API isn't enabled.

### The fix

Switch the fallback from the People API to the OAuth userinfo endpoint (`oauth2/v2/userinfo`). This endpoint:
- Only requires the `userinfo.profile` scope (no separately-enabled API)
- Returns `{"name": "...", ...}` at the top level (simpler response shape)
- Is the same endpoint the CLI already uses for email lookup after login (`auth_commands.rs`)

### Other improvements

- **Diagnostic for empty display name:** When the userinfo endpoint succeeds but returns no `name` field (e.g. a freshly-provisioned account with no profile configured), a note is now printed to stderr. Previously this was completely silent — the user got no display name with no explanation.
- **Context-appropriate tips:** `GOOGLE_WORKSPACE_CLI_TOKEN` users now see a tip about setting a token with the profile scope, instead of being told to run `gws auth login` (which doesn't apply to them).
- **Function renames:** `fetch_profile_display_name` → `fetch_userinfo_display_name`, `parse_profile_display_name` → `parse_userinfo_display_name` to reflect the new endpoint.

### What this does NOT change

The primary display name path (sendAs identity lookup via `/users/me/settings/sendAs`) is untouched. This fix only affects the fallback for Workspace accounts where sendAs returns an empty `displayName`. The token cache and auth flow are also unchanged.

### Verified

- Built the fixed binary and created a draft via `+send --draft` — From header correctly shows display name, no spurious tip printed.
- 703 tests pass, clippy clean.

Fixes #644

## Checklist:

### DIFF
--- a/.changeset/fix-gmail-display-name-fallback.md
+++ b/.changeset/fix-gmail-display-name-fallback.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Use the OAuth userinfo endpoint for Gmail sender display-name fallback instead of the People API, which required the API to be separately enabled in the GCP project.
--- a/crates/google-workspace-cli/src/auth_commands.rs
+++ b/crates/google-workspace-cli/src/auth_commands.rs
@@ -600,8 +600,9 @@ async fn handle_login_inner(
 
     // Ensure openid + email + profile scopes are always present so we can
     // identify the user via the userinfo endpoint after login, and so the
-    // Gmail helpers can fall back to the People API to populate the From
-    // display name when the send-as identity lacks one (Workspace accounts).
+    // Gmail helpers can fall back to the OAuth userinfo endpoint to populate
+    // the From display name when the send-as identity lacks one (Workspace
+    // accounts).
     let identity_scopes = [
         "openid",
         "https://www.googleapis.com/auth/userinfo.email",
--- a/crates/google-workspace-cli/src/helpers/gmail/mod.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/mod.rs
@@ -552,6 +552,18 @@ fn resolve_sender_from_identities(
     }
 }
 
+fn uses_env_access_token() -> bool {
+    std::env::var_os("GOOGLE_WORKSPACE_CLI_TOKEN").is_some_and(|token| !token.is_empty())
+}
+
+fn profile_scope_tip(using_env_token: bool) -> &'static str {
+    if using_env_token {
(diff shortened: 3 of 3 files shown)


malob:
Bump to keep open — awaiting maintainer review.

PR #693, opened 2026-04-09
fix(gmail): handle message headers case-insensitively

### LINKED ISSUE #642: bug(gmail): parse_message_headers uses case-sensitive match, drops CC/headers with non-canonical casing
## Description

## Reproduction

2. Run:
   ```bash
   ```

## Root Cause

```rust
match name {
    "From" => ...
    "Reply-To" => ...
    "To" => ...
}
```

## Scope

| Header | Impact if missed |
|--------|-----------------|

## Internal Inconsistency

```rust
    ...
    ...
}
```

## Suggested Fix

```rust
    "from" => ...
    "reply-to" => ...
    "to" => ...
    "cc" => ...
    "subject" => ...
    "date" => ...
    "message-id" => ...
    "references" => ...
    _ => {}
}
```

## Related

## Environment

- gws 0.22.3
- macOS (arm64)

jeevan6996:
## Summary
- update parse_message_headers to match header names case-insensitively, aligning behavior with RFC semantics and existing helper logic
- ensure reply/reply-all keeps recipients and threading metadata even when headers arrive as FROM/TO/CC or mixed-case variants
- add a regression test that exercises mixed-case From/Reply-To/To/Cc/Date/Message-ID/References headers
- include a patch changeset for release automation

Fixes #642

### DIFF
--- a/.changeset/gmail-header-case-insensitive.md
+++ b/.changeset/gmail-header-case-insensitive.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Fix Gmail reply header parsing to treat message header names as case-insensitive so recipients and threading metadata are preserved.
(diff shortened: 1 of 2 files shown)


jeevan6996:
Update: I have now signed the Google CLA. The cla/google check is still showing the previous failed run; I cannot rerun workflows on this repo from my permissions. Could a maintainer trigger a rerun if needed, or confirm once the CLA bot refreshes status? Thanks.

jeevan6996:
Quick check-in: this PR appears clean aside from the lingering cla/google pending state from an older run. I have already signed the Google CLA. Could a maintainer trigger a checks rerun or refresh the CLA status if needed?

PR #700, opened 2026-04-11
fix(tasks): warn when due time is ignored by API

### LINKED ISSUE #696: Bug - Tasks - set due time
When I try to set the due date through the `gws-cli` by defining both the day and the time, it sets just the date skipping the time.

jeevan6996:
## Summary
- add a targeted warning when requests to Google Tasks include a non-midnight `due` timestamp, since the Tasks API stores only date information and discards time-of-day
- scope the warning to Tasks write operations (`POST`/`PUT`/`PATCH`) so other APIs are unaffected
- add unit tests for non-midnight detection, midnight-no-warning behavior, and non-Tasks API behavior
- include a patch changeset

Fixes #696

### DIFF
--- a/.changeset/tasks-due-time-warning.md
+++ b/.changeset/tasks-due-time-warning.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Warn when a Google Tasks request includes a non-midnight `due` time, since the API stores only the date and ignores time-of-day.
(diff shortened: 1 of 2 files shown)


jeevan6996:
Addressed the review concern about terminal safety: warning output now sanitizes the due timestamp before printing to stderr (`sanitize_for_terminal`). Pushed in 83df826.

jeevan6996:
Addressed the new review point as well: `extract_due_string_with_non_midnight_time` now detects non-midnight fractional seconds (e.g. 00:00:00.500Z) by scanning the time segment for any non-zero digit, and includes a regression test for this case. Pushed in 1271b41.

PR #736, opened 2026-04-20
fix(auth): map meet service to meetings scope prefix

### LINKED ISSUE #556: Auth login missing scopes for People and Meet APIs
## Summary

`gws auth login` does not offer OAuth scopes for the People API (`contacts.readonly`) or Meet REST API (`meetings.space.created`) in its interactive scope selector, even though gws supports both `people` and `meet` as services with full API discovery.

## Steps to reproduce

1. Run `gws auth login`
2. Observe the interactive scope selector — only 9 scopes offered:
   - Drive, Sheets, Gmail, Calendar, Docs, Slides, Tasks, Pub/Sub, Cloud Platform
3. Run `gws schema people.people.searchContacts` — shows it requires `contacts` or `contacts.readonly` scope
4. Run `gws schema meet.spaces.create` — shows it requires `meetings.space.created` scope
5. Try to call `gws people people searchContacts ...` — fails with "insufficient authentication scopes"
6. Try to call `gws meet spaces create` — fails with "insufficient authentication scopes"

## Expected behavior

The auth login scope selector should include all scopes needed by the services gws supports, including:
- `https://www.googleapis.com/auth/contacts.readonly` (or `contacts`) for the People API
- `https://www.googleapis.com/auth/meetings.space.created` for the Meet API

## Workaround attempted

Manually editing the OAuth authorization URL to include the missing scopes before submitting — Google's consent screen shows the additional permissions, but gws's local callback server does not capture them in the resulting credential.

## Environment

- gws 0.13.3
- Linux x86_64

## Context

We're building an MCP server on top of gws ([google-workspace-mcp](https://github.com/aaronsb/google-workspace-mcp)) and discovered this while expanding coverage to all gws-supported services. The factory architecture generates tools from gws's own API discovery, so it's natural to expect auth to cover the same services.

jbasso-dev:
## Description

`gws auth login -s meet` (and multi-service filters like `-s drive,gmail,calendar,sheets,docs,meet,tasks`) silently requests zero Meet scopes, so subsequent calls like `gws meet conferenceRecords list` fail with `403 insufficient authentication scopes`.

The cause is a naming mismatch: the CLI service alias is `meet`, but Google's Meet API scopes are `https://www.googleapis.com/auth/meetings.*`. `map_service_to_scope_prefixes` returned `"meet"` via its catch-all arm, so `scope_matches_service` filtered every Meet scope out of both the picker and `find_unmatched_services`.

The fix is a one-liner in the same spot that already special-cases `sheets`, `slides`, `docs`, and `people`:

```rust
"meet" => vec!["meetings"],
```

Once the filter correctly classifies `meet` as unmatched, the existing `augment_with_dynamic_scopes` path picks up all three Meet scopes from the Discovery document on its own. No changes to `FULL_SCOPES` or `SCOPE_ENTRIES` — those intentionally list only the core Workspace services, not every supported API.

Closes #556. Continuation of #565, which was auto-closed by the stale-bot. Credit to @anshul-garg27 for the original diagnosis; this PR is narrower (just the mapping change) and aligned with the convention set by #414.

## Checklist:

- [x] I have run `cargo clippy -- -D warnings` and resolved all warnings. (13 pre-existing errors in `main.rs` reproduce on clean `main` — out of scope.)
- [x] I have provided a Changeset file (e.g. via \`pnpx changeset\`) to document my changes.

### DIFF
--- a/.changeset/auth-meet-scopes.md
+++ b/.changeset/auth-meet-scopes.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Fix `gws auth login -s meet` not granting any Meet scopes. The Meet API exposes scopes under the `meetings.*` prefix (e.g. `meetings.space.readonly`), but the CLI's service alias is `meet`, so the scope-picker filter matched nothing and silently dropped every Meet scope. Map the `meet` service to the `meetings` scope prefix so the existing dynamic-augmentation path pulls all Meet scopes from the Discovery document when `-s meet` is supplied.
--- a/crates/google-workspace-cli/src/auth_commands.rs
+++ b/crates/google-workspace-cli/src/auth_commands.rs
@@ -841,6 +841,7 @@ fn map_service_to_scope_prefixes(service: &str) -> Vec<&str> {
         "slides" => vec!["presentations"],
         "docs" => vec!["documents"],
         "people" => vec!["contacts", "directory"],
+        "meet" => vec!["meetings"],
         s => vec![s],
     }
 }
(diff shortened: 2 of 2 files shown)


jbasso-dev:
Reopening — this was auto-closed by the stale bot after 72h of inactivity, but the fix is still valid and #556 is still open. The change is a one-liner mapping `meet` → `meetings` scope prefix (matching the existing convention for `sheets`, `slides`, `docs`, `people`) with tests. Happy to address any review feedback.

PR #744, opened 2026-04-22
fix(windows): force UTF-8 console output at startup

### LINKED ISSUE #742: Non-ASCII characters in API responses are mangled on Windows when the console codepage is CP-1252 (default)
## Body

**Environment:**
- OS: Windows 10/11
- Shell / invocation: `cmd /c gws ...` (also reproducible from a plain `cmd.exe` prompt)
- Default console codepage: CP-1252 (en-US, "Windows-1252")

---

## Summary

Non-ASCII UTF-8 bytes in API response data are emitted to stdout as if each byte were a CP-1252 character. The classic `â€™` and `â€"` patterns appear in place of smart quotes and em-dashes respectively.

This is the standard "UTF-8 bytes interpreted as CP-1252" mojibake signature. The underlying API response is correct UTF-8; the corruption is introduced in `gws`'s output pipeline on Windows when the console codepage is not UTF-8.

## Reproduction

Any `gws` command whose response contains smart punctuation or other non-ASCII characters will reproduce this. A reliable test case:

```
gws docs documents get --document-id <ID> --fields "body(content(paragraph(elements(textRun(content)))))"
```

on a Google Doc containing AI-generated prose (which typically includes smart quotes and em-dashes).

**Expected:**
```
it's a balanced — considered — approach
```

**Actual:**
```
itâ€™s a balanced â€" considered â€" approach
```

**Confirmation the source data is correct:**
1. Open the same document in the Google Docs web UI — characters display correctly.
2. Fetch the same document via any UTF-8-aware HTTP client — characters are correct UTF-8.
3. Run the same `gws` command after executing `chcp 65001` in the same shell session — output is correct.

### `chcp 65001` — works only for persistent sessions

```
chcp 65001
gws docs documents get ...
```

This resolves the issue for interactive sessions. However, any automation layer that spawns a fresh `cmd /c` per command gets a new child process with the default codepage. The `chcp` change does not persist across `cmd /c` invocations.

### PowerShell with console encoding set

```
pwsh -Command "[Console]::OutputEncoding = [System.Text.Encoding]::UTF8; gws docs documents get ..."
```

Resolves the mojibake in some scenarios, but introduces different reliability issues in automated environments (stdout capture failures, timeouts). Not a clean substitute for a fix in `gws` itself.

## Hypothesis

On Windows, Rust's default stdout writer honors the console's output codepage. When the codepage is CP-1252 and the program writes UTF-8 bytes, each byte above 0x7F is re-rendered through the CP-1252 glyph table. `gws` should either:

- Call `SetConsoleOutputCP(65001)` at startup (via `windows-sys` or equivalent), or
- Write to stdout through a binary-mode handle that bypasses codepage translation entirely.

Both approaches are well-established for Rust CLIs targeting Windows and should be wrapped in `#[cfg(target_os = "windows")]` to leave Unix builds unaffected.

## Impact

Affects every read operation whose response includes non-ASCII characters:

- Google Docs content (smart quotes and em-dashes are common in AI-assisted writing)
- Gmail message bodies (smart punctuation in signatures, quoted replies, non-English correspondence)
- Contact names with accented characters
- Calendar event titles and descriptions with non-ASCII characters
- File names containing non-ASCII characters

The corruption is silent — no error, no warning, exit 0. Downstream consumers that store or forward `gws` output will persist the corrupted bytes.

## Requested behavior

On Windows, force UTF-8 output regardless of the user's console codepage. Two concrete options:

1. **Call `SetConsoleOutputCP(65001)` at startup.** Minor downside: modifies global console state, which could surprise users with a specific codepage set for other tools.
2. **Write stdout as raw bytes, bypassing codepage translation.** Use `GetStdHandle(STD_OUTPUT_HANDLE)` + `WriteFile`, or write to `std::io::stdout().lock()` after confirming the underlying handle mode. More correct long-term — does not modify global state.

Either approach should be conditional on `cfg(target_os = "windows")`.

hoyt-harness:
## Summary

Fixes #742.

On Windows, `gws` inherits the system console codepage (typically CP-1252 or CP-437). When Rust's `println!` writes UTF-8 bytes, the console interprets them using the active codepage, producing mojibake for any non-ASCII characters in API responses (e.g. smart quotes, accented names, emoji).

**Fix:** Call `SetConsoleOutputCP(65001)` via a thin inline FFI shim at the very start of `main()`, before any output is produced. This tells the Windows Console to interpret subsequent output as UTF-8. The call is gated behind `#[cfg(windows)]` so non-Windows builds are unaffected.

No new dependency is added — the function is part of `kernel32.dll` and is called directly via an `extern "system"` block.

### Files changed

| File | Change |
|---|---|
| `crates/google-workspace-cli/src/output.rs` | Add `set_console_utf8()` + Windows-conditional test |
| `crates/google-workspace-cli/src/main.rs` | Call `set_console_utf8()` at startup |
| `crates/google-workspace-cli/src/helpers/script.rs` | Collapse nested `if`-in-`match` to satisfy `clippy::collapsible_match` on latest stable (lint trigger on Rust 1.87+; required for `cargo clippy -- -D warnings` to pass locally) |
| `.changeset/fix-windows-utf8-console-codepage.md` | Patch changeset |

The `script.rs` change is a ride-along lint fix — happy to split it into a separate PR if preferred.

## Test plan

- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo test` passes (702 tests pass; 2 pre-existing Windows-only failures in `auth::tests::test_get_quota_project_*` — those tests set `HOME` but `dirs::home_dir()` on Windows uses `USERPROFILE`, unrelated to this PR)
- [x] `set_console_utf8_sets_codepage_65001` test: calls `SetConsoleOutputCP(65001)` and then `GetConsoleOutputCP()` — confirmed returns `65001` on Windows (skips assertion gracefully when no console is attached, e.g. in CI)
- [x] Binary built and smoke-tested on Windows 11 x86_64

### DIFF
--- a/.changeset/fix-windows-utf8-console-codepage.md
+++ b/.changeset/fix-windows-utf8-console-codepage.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Force UTF-8 console output on Windows by calling `SetConsoleOutputCP(65001)` at startup, preventing mojibake when the system default codepage is CP-1252
--- a/crates/google-workspace-cli/src/helpers/script.rs
+++ b/crates/google-workspace-cli/src/helpers/script.rs
@@ -169,13 +169,8 @@ fn process_file(path: &Path) -> Result<Option<serde_json::Value>, GwsError> {
             filename.trim_end_matches(".js").trim_end_matches(".gs"),
         ),
         "html" => ("HTML", filename.trim_end_matches(".html")),
-        "json" => {
-            if filename == "appsscript.json" {
-                ("JSON", "appsscript")
-            } else {
-                return Ok(None);
-            }
-        }
+        "json" if filename == "appsscript.json" => ("JSON", "appsscript"),
+        "json" => return Ok(None),
         _ => return Ok(None),
     };
 
--- a/crates/google-workspace-cli/src/main.rs
+++ b/crates/google-workspace-cli/src/main.rs
@@ -47,6 +47,10 @@ use error::{print_error_json, GwsError};
 
 #[tokio::main]
 async fn main() {
+    // Force UTF-8 console output on Windows; no-op on other platforms.
+    #[cfg(windows)]
+    output::set_console_utf8();
+
     // Load .env file if present (silently ignored if missing)
     let _ = dotenvy::dotenv();
 
--- a/crates/google-workspace-cli/src/output.rs
+++ b/crates/google-workspace-cli/src/output.rs
@@ -86,6 +86,27 @@ pub(crate) fn info(msg: &str) {
     eprintln!("{}", sanitize_for_terminal(msg));
 }
 
+// ── Windows console codepage ───────────────────────────────────────────
+
+/// Force the Windows console output codepage to UTF-8 (CP 65001) at startup.
+/// Without this, the default system codepage (typically CP-1252) mangles
+/// non-ASCII characters printed by `println!` before Rust's stdio layer can
+/// re-encode them.
+///
+/// Failure is non-fatal — terminals that already use UTF-8 will behave
+/// correctly regardless, and the call is safe to make unconditionally on any
+/// Windows version that ships `kernel32.dll`.
+#[cfg(windows)]
+pub(crate) fn set_console_utf8() {
+    extern "system" {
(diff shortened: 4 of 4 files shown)


hoyt-harness:

Issue #741, opened 2026-04-21
`auth login -s` / `--services` silently drops unknown service names and injects `cloud-platform` scope without documentation


hoyt-harness:
## Body

**Version:** `gws` v0.22.5 (Windows x86_64 binary from GitHub Releases)

**Environment:**
- OS: Windows 10/11
- Auth flow: OAuth 2.0 desktop client, consent screen in External / Testing mode

---

## Summary

`gws auth login -s <list>` is a service-picker filter that accepts a comma-separated list of service names and resolves each to a default scope. In practice, the flag has two undocumented behaviors:

1. **Unknown names in the list are silently dropped.** No warning, no error. The surviving names are granted; the dropped names are not.
2. **An undocumented `cloud-platform` scope is injected** whenever any of the surviving services have a sanitization layer dependency. This happens even when the user explicitly omitted `cloud-platform` from their intent.

The combined effect: a user who runs a single `auth login -s` command gets a token containing a scope set materially different from what they typed, with no diagnostic indicating the mismatch.

### Case A: Silent-drop of unknown names

```
gws auth login -s gmail,drive,calendar,calendar.events,documents,spreadsheets,tasks,contacts,contacts.other.readonly
```

**Expected:** Grant all nine named scopes. The help text example (`-s drive,gmail,sheets`) suggests the flag takes service names, and several items in the list (`calendar.events`, `contacts.other.readonly`) are valid scope URI suffixes, so a user could reasonably interpret the flag as accepting either form.

**Actual:** After browser consent, `gws auth status` shows the valid service aliases each resolved to one default scope — plus `cloud-platform` which was not in the list. The names `contacts` and `contacts.other.readonly` were silently dropped. No stderr warning. Exit code 0.

**Cross-check:** Running `gws contacts --help` after the re-auth fails with a `Known services: ...` error — so the CLI has the data to validate service names at parse time but does not use it during `auth login -s`.

### Case B: Undocumented `cloud-platform` injection

Same command as Case A. `gws auth status` shows `cloud-platform` in the granted-scope list even though it was not requested. No step of the `auth login -s` flow surfaced this addition.

**Working path (no injection):** Deleting `token_cache.json` and re-running with `--scopes` (plural) using explicit scope URIs:

```
gws auth login --scopes https://www.googleapis.com/auth/gmail.modify,https://www.googleapis.com/auth/drive,...
```

This produces a clean grant of exactly the intended scopes with no `cloud-platform` added.

## Hypothesis

[...]

PR #747, opened 2026-04-22
fix(auth): validate --services names and remove automatic cloud-platform injection

hoyt-harness:
## Summary

Fixes two related bugs in `gws auth login -s` / `--services` reported in #741:

- **Unknown service names were silently dropped.** `parse_login_args()` now validates every token in the `-s` list against the built-in service registry (`crate::services::SERVICES`) before opening the browser. If any token is unrecognised, the command exits non-zero with a clear error that names the bad token(s) and lists all valid service names. Tokens are lowercased before checking, so `-s Drive` and `-s drive` both work.

- **`cloud-platform` was unconditionally injected when a services filter was active.** `scope_matches_service()` previously returned `true` for `cloud-platform` regardless of what services the user asked for, so `-s drive` silently included `cloud-platform` in the resulting token. The unconditional pass-through is removed; `cloud-platform` is now filtered the same way as any other scope. `run_discovery_scope_picker()` retains the auto-add behaviour when *no* services filter is active (the existing unfiltered-interactive flow is unchanged).

**Behaviour change note:** Users who relied on `-s <anything>` always producing a token that includes `cloud-platform` will need to add `cloud-platform` explicitly via `--scopes` or omit the `-s` filter. This is an intentional correction: the previous injection violated the principle of least privilege and contradicted explicitly declined scopes.

Also fixes a pre-existing `clippy::collapsible_match` in `helpers/script.rs` (same lint fixed in #744, which targets the same upstream/main base).

## Test plan

- [x] `parse_login_args_rejects_unknown_service` — error returned for typo token, message names the bad token and lists valid services
- [x] `parse_login_args_accepts_known_services` — `Drive,Gmail` accepted and lowercased correctly
- [x] `scope_matches_service_cloud_platform_does_not_match_unrelated_service` — cloud-platform no longer auto-passes the drive filter
- [x] `resolve_scopes_with_services_filter` — filtered result contains only drive/gmail scopes, no cloud-platform
- [x] `resolve_scopes_services_takes_priority_with_readonly` — readonly + drive filter: only drive scopes
- [x] `resolve_scopes_services_takes_priority_with_full` — full + gmail filter: only gmail scopes
- [x] Full test suite: `cargo test -p google-workspace-cli` — 701 passed, 2 pre-existing failures (unrelated to this change), 0 new failures

### DIFF
--- a/.changeset/fix-741-auth-services-validation.md
+++ b/.changeset/fix-741-auth-services-validation.md
@@ -1,0 +1,7 @@
+---
+"@googleworkspace/cli": patch
+---
+
+`auth login -s` / `--services`: reject unknown service names at parse time with a clear error listing valid names. Previously, unrecognised tokens were silently dropped, causing a token to cover fewer services than the user intended.
+
+**Behavior change:** `cloud-platform` is no longer injected automatically when a `--services` filter is active. It was previously added to every filtered login regardless of the services requested. It still appears when using `--full` (no filter) or the interactive discovery picker (no filter).
--- a/crates/google-workspace-cli/src/auth_commands.rs
+++ b/crates/google-workspace-cli/src/auth_commands.rs
@@ -448,7 +448,7 @@ pub async fn handle_auth_command(args: &[String]) -> Result<(), GwsError> {
 
     match matches.subcommand() {
         Some(("login", sub_m)) => {
-            let (scope_mode, services_filter) = parse_login_args(sub_m);
+            let (scope_mode, services_filter) = parse_login_args(sub_m)?;
 
             handle_login_inner(scope_mode, services_filter).await
         }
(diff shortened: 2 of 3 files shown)


hoyt-harness:
Thanks for the review feedback. Both issues addressed in the latest commit:

1. **Platform scope allowlist** — Added `PLATFORM_SCOPES = ["cloud-platform", "pubsub"]` to the validation predicate and the valid-names list in the error message. These scopes exist in `SCOPE_ENTRIES`/`FULL_SCOPES` but have no entry in the Workspace service registry (`SERVICES`), so the original check incorrectly rejected them. Since this PR also removes the automatic `cloud-platform` injection, without this fix users would have had no way to request that scope when using `-s`.

2. **Escape sequence injection** — Applied `.escape_debug()` on unknown tokens before including them in the error string.

New test: `parse_login_args_accepts_platform_scopes` covers the accept path for both tokens.

hoyt-harness:

PR #748, opened 2026-04-22
fix(drive): pass supportsAllDrives=true in +upload to support Shared Drives

### LINKED ISSUE #722: gws drive +upload doesn't work with Shared Drive folders
Hello together,
something I've stumbled upon while using the workspace cli:

## What's wrong

  Trying to upload into a Shared Drive folder fails with `404 File not found`,
  because `+upload` doesn't pass `supportsAllDrives: true` to the Drive API.
  There's also no way to add it manually — the helper rejects `--params`.

  So the helper is effectively unusable for anyone whose Drive content lives
  on a Shared Drive (which is most teams).

  This is related to #327 (which is about telling agents to add the flag when
  calling the raw API), but separate: that issue is about awareness, this one
  is about the helper having no way to set the flag at all.

  ## How to reproduce

  CLI version: `@googleworkspace/cli` v0.22.5

  ```bash
  echo "test" > /tmp/test.txt
  gws drive +upload /tmp/test.txt --parent <SHARED_DRIVE_FOLDER_ID>
  ```

  Result:
  ```
  {
    "error": {
      "code": 404,
      "message": "File not found: <SHARED_DRIVE_FOLDER_ID>.",
      "reason": "notFound"
    }
  }
  ```

  Trying to add the flag manually:
  ```bash
  gws drive +upload /tmp/test.txt --parent <FOLDER_ID> \
    --params '{"supportsAllDrives": true}'
  ```
  also fails:
  ```
  error[validation]: error: unexpected argument '--params' found
  ```

  ## What works

  Skip the helper, use the raw API:
  ```bash
  gws drive files create \
    --params '{"supportsAllDrives": true}' \
    --json '{"name": "test.txt", "parents": ["<FOLDER_ID>"]}' \
    --upload /tmp/test.txt
  ```

  ## Suggested fixes (any one would resolve)

  1. **Set `supportsAllDrives: true` by default in `+upload`.** The flag is
     safe for non-shared drives — it has no effect when the parent isn't on
     a Shared Drive — so passing it unconditionally costs nothing and fixes
     the helper.
  2. **Add a `--shared-drive` boolean flag** that opts in.
  3. **Allow `--params` as an escape hatch on `+upload`** (and other helpers,
     for consistency with the raw API surface).

  Option 1 is least friction for users (they don't need to know Shared Drives
  are a thing); option 3 is most consistent with the rest of the CLI.

  ## Environment

  - OS: Windows 10 (Git Bash)
  - gws CLI: 0.22.5

________________________________________

Cheers
Dan

hoyt-harness:
## Problem

`drive +upload` calls `files.create` without the `supportsAllDrives=true` query parameter. The Drive API rejects any request that targets a Shared Drive folder (via `--parent`) unless this parameter is present, returning a 404 or 403 even when the user has full access to the Shared Drive.

Reported in #722.

## Fix

Pass `supportsAllDrives: true` unconditionally in the query params for every `+upload` call. This parameter is documented as harmless for My Drive uploads — it simply opts the request into Shared Drive awareness.

```
$ gws drive +upload ./file.txt --parent SHARED_DRIVE_FOLDER_ID --dry-run
{
  "query_params": [["supportsAllDrives", "true"]],
  ...
}
```

Updated `after_help` text to document Shared Drive support.

## Test plan

- [x] Dry-run output confirms `supportsAllDrives=true` appears in `query_params`
- [x] All 5 existing `helpers::drive` unit tests pass
- [x] My Drive uploads unaffected (parameter is a no-op when not targeting a Shared Drive)

Closes #722

### DIFF
--- a/.changeset/fix-722-drive-upload-shared-drives.md
+++ b/.changeset/fix-722-drive-upload-shared-drives.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+`drive +upload`: pass `supportsAllDrives=true` on every upload request so files can be placed inside Shared Drive folders via `--parent`. Previously, any `--parent` pointing to a Shared Drive folder was rejected by the API with a 404 or permission error even when the user had access.
--- a/crates/google-workspace-cli/src/helpers/drive.rs
+++ b/crates/google-workspace-cli/src/helpers/drive.rs
@@ -56,11 +56,13 @@ impl Helper for DriveHelper {
 EXAMPLES:
   gws drive +upload ./report.pdf
   gws drive +upload ./report.pdf --parent FOLDER_ID
+  gws drive +upload ./report.pdf --parent SHARED_DRIVE_FOLDER_ID
   gws drive +upload ./data.csv --name 'Sales Data.csv'
 
 TIPS:
   MIME type is detected automatically.
-  Filename is inferred from the local path unless --name is given.",
+  Filename is inferred from the local path unless --name is given.
+  Shared Drive folders are supported: pass the folder ID via --parent.",
                 ),
         );
         cmd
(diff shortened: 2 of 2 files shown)


hoyt-harness:

PR #750, opened 2026-04-22
feat(executor): stream binary responses to stdout when --output is omitted

### LINKED ISSUE #743: File-path handling is inconsistent across flags and endpoints: cwd restrictions on `--output`/`--upload`, `files export` writes to `download.txt` instead of stdout, `events delete` creates spurious `download.html` on 204

hoyt-harness:
## Summary

`drive files export`, `drive files get --alt media`, and any other binary endpoint previously always wrote content to a `download.{ext}` file in cwd and printed a status JSON object to stdout, with no way to pipe the content to another command without an intermediate file.

This PR changes the default to the Unix/curl convention:

| Invocation | Before | After |
|---|---|---|
| `gws drive files export ...` | writes `download.txt` to cwd; prints status JSON | raw bytes on stdout |
| `gws drive files export ... --output file.txt` | writes `file.txt`; prints status JSON | writes `file.txt`; prints status JSON (**unchanged**) |
| `gws drive files get ... --alt media` | writes `download.pdf` to cwd; prints status JSON | raw bytes on stdout |
| `gws drive files get ... --alt media --output file.pdf` | writes `file.pdf`; prints status JSON | writes `file.pdf`; prints status JSON (**unchanged**) |

The `mime_to_extension()` helper, which existed only to derive the `download.{ext}` fallback filename, is removed as dead code.

> **Note:** This PR addresses the 4b sub-issue of #743. The 4d sub-issue (calendar events delete 204 mis-routing) is a separate, independent fix in PR #749.

## Breaking change

Scripts that rely on a `download.txt` / `download.pdf` / etc. appearing in cwd after `drive files export` or `drive files get --alt media` must be updated to either:

```bash
gws drive files export ... > output.txt        # redirect stdout
gws drive files export ... --output output.txt # explicit --output flag
```

## Test plan

- [ ] `gws drive files export --file-id <ID> --params '{"mimeType":"text/plain"}' | head` streams content to stdout without creating any file in cwd
- [ ] `gws drive files export --file-id <ID> --params '{"mimeType":"text/plain"}' --output export.txt` saves `export.txt` and prints status JSON (existing behaviour)
- [ ] `gws drive files get --file-id <ID> --params '{"alt":"media"}' > file.pdf` produces a valid PDF
- [ ] `cargo clippy -- -D warnings` passes
- [ ] `cargo test` passes (excluding the two pre-existing `auth::tests::test_get_quota_project_*` failures that are environment-specific ADC file reads unrelated to this change)

Closes #743 (sub-issue 4b)

### DIFF
--- a/.changeset/fix-binary-stream-to-stdout.md
+++ b/.changeset/fix-binary-stream-to-stdout.md
@@ -1,0 +1,20 @@
+---
+"@googleworkspace/cli": minor
+---
+
+Binary responses now stream to stdout when no `--output` flag is provided.
+
+Previously, commands like `drive files export` and `drive files get --alt media`
+always wrote content to a `download.{ext}` file in the current working directory
+and printed a status JSON object to stdout, with no way to pipe the content
+directly to another process.
+
+Now, omitting `--output` streams the raw bytes to stdout (the Unix/curl default).
+Use `--output <path>` to save the content to a named file as before.
+
+The `mime_to_extension` helper, which existed only to generate the default
+`download.{ext}` filename, has been removed as it is no longer needed.
+
+**Migration note:** Scripts that read `download.txt` / `download.pdf` / etc.
+from cwd after running `drive files export` must be updated to either redirect
+stdout (`gws ... > file.txt`) or supply `--output file.txt` explicitly.
--- a/crates/google-workspace-cli/src/executor.rs
+++ b/crates/google-workspace-cli/src/executor.rs
@@ -336,7 +336,11 @@ async fn handle_json_response(
     Ok(false)
 }
 
-/// Handle a binary response by streaming it to a file.
+/// Handle a binary response by streaming it to a file or stdout.
+///
+/// When `output_path` is provided the content is saved to that file and a
+/// status JSON is printed/captured.  When no path is given the raw bytes are
+/// streamed directly to stdout so the caller can pipe or redirect them.
 async fn handle_binary_response(
     response: reqwest::Response,
     content_type: &str,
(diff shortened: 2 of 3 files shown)


hoyt-harness:
Good catch — two issues fixed in the latest commit:

1. **`capture_output` regression** — the stdout path now respects the flag. When `capture_output` is true the stream is drained (bytes counted) without writing to stdout, and the function returns `{status, mimeType, bytes}` metadata consistent with the file-output path. When false it streams to stdout as before.

2. **`total_bytes` tracking** — added to the stdout path so both code paths return consistent metadata.

Note: I kept the `helpers/script.rs` clippy fix in this PR since `upstream/main` carries that pre-existing lint error — removing it would cause this PR's clippy CI check to fail.

hoyt-harness:
**Line 397 (TTY guard) — fixed.** This is a real gap introduced by the PR: pre-change, binary always went to a file; post-change it can reach a terminal. Added an `IsTerminal` check (stable since Rust 1.70; CI uses `dtolnay/rust-toolchain@stable` so no MSRV concern, no new dep):

```rust
if !capture_output && std::io::IsTerminal::is_terminal(&std::io::stdout()) {
    return Err(GwsError::Validation(
        "Refusing to stream binary content to a terminal. \
         Use --output <path> to save to a file, or redirect stdout (e.g. '> file')."
            .to_string(),
    ));
}
```

The `!capture_output` guard ensures the check only fires when we'd actually write bytes to stdout — the capture path drains the stream without writing, so no TTY risk there.

---

**Line 369 (skip file write when `capture_output=true`) — respectfully declining.** The two branches are asymmetric by design, not by accident:

- The **stdout path** skips writing because there's no separate destination — writing bytes to stdout and returning a captured `Value` are mutually exclusive.
- The **file path** has two independent destinations: the disk (honoring the user's explicit `--output` request) and the return value. Skipping the disk write when `capture_output=true` would silently ignore a user-specified argument, which is the worse behaviour. The existing pattern — write the file, then either print or return the metadata depending on `capture_output` — is intentional and consistent with how the JSON response path works.

hoyt-harness:

PR #751, opened 2026-04-22
fix(gmail): rename +read --format to --body-format to avoid global flag collision

### LINKED ISSUE #740: Multiple commands return exit 0 with empty stdout when output serialization fails, silently swallowing successful API responses
## Body

**Environment:**
- OS: Windows 10/11
- Auth: OAuth user token via `gws auth login --scopes <explicit URI list>`
- Shared Drive in use: Yes

---

## Summary

Several `gws` commands exit with code 0 and no stderr error, but write nothing to stdout, on API calls that should return data. Cross-checks against the same Workspace account via Google's APIs and via alternative `gws` command forms confirm the underlying data exists — the response is being produced by the API and then silently dropped by `gws` before reaching stdout.

This is distinct from legitimate empty-result cases (where `gws` correctly writes `{}` or an empty items array).

## Affected endpoints (confirmed)

All reproduced against the same authenticated session.

| Endpoint | Symptom | Cross-check confirming data exists |
|---|---|---|
| `gws people people connections list` | empty stdout, exit 0 | Account has contacts — confirmed via `people contactGroups get` with `maxMembers` returning populated `memberResourceNames` |
| `gws people people searchContacts` | empty stdout, exit 0 | `searchContacts` with a known contact email returns empty even though the contact exists |
| `gws gmail users labels list` | empty stdout, exit 0 | 8 user labels exist (verified independently) |
| `gws gmail +read --message-id <ID>` | empty stdout, exit 0; stderr notes `"unknown output format 'text', falling back to json"` but nothing follows | Same message ID returns full content via `gws gmail users messages get` |
| `gws gmail users messages get` for a specific message ID | empty stdout, exit 0 | Same command with a different message ID returns the full multipart message correctly |

### Case A: `gmail users labels list`

```
gws gmail users labels list --params "{\"userId\":\"me\"}"
```

**Expected:** JSON object containing `labels` array with all user and system labels.

**Actual:** Exit 0. Empty stdout. Stderr shows only the cosmetic `Using keyring backend: keyring` line.

### Case B: `gmail users messages get` — ID-specific

```
gws gmail users messages get --params "{\"userId\":\"me\",\"id\":\"<id>\",\"format\":\"full\"}"
```

**Expected:** Full message resource (headers, body parts, snippet).

**Actual:** Exit 0. Empty stdout.

**Contrast:** The same command with a different message ID (from the same account, same token) returns a complete multipart message resource. The only difference is the message ID.

### Case C: `gmail +read`

```
gws gmail +read --message-id <id>
```

**Expected:** Message text rendered to stdout.

**Actual:** Exit 0. Empty stdout. Stderr includes `unknown output format 'text', falling back to json` — but no JSON follows.

### Case D: `people people searchContacts`

```
gws people people searchContacts --params "{\"query\":\"<term>\",\"readMask\":\"names,emailAddresses\"}"
```

**Expected:** JSON with search results matching known contacts.

**Actual:** Exit 0. Empty stdout.

## Hypothesis

The failure mode looks like a swallowed serialization error. Evidence:

1. The `+read` case emits `unknown output format 'text', falling back to json` to stderr before producing empty stdout — the output pipeline is entering a fallback path that then fails silently.
2. The `messages get` case is ID-specific, not command-specific — which suggests sensitivity to response shape (e.g., specific multipart structures, header configurations, or MIME types present in one message but not another).
3. The broken endpoints share a common feature: they return variable-shape payloads. `labels list` returns a heterogeneous list (system and user labels have different field sets). `connections list` and `searchContacts` return `Person` objects with optional fields whose presence depends on the contact. Meanwhile, the endpoints that work reliably return shapes that are stable per-call.

A reasonable guess: the output formatter panics or errors on certain response shapes and the error is caught and discarded (e.g., via `.ok()` or `let _ = ...`) rather than surfaced.

## Impact

**Critical for automation.** Any script, skill, or agent that treats `exit 0 + empty stdout` as "no results" will:

- Miss Gmail labels entirely on a fresh read
- Fail to retrieve certain messages
- Skip contact searches entirely
- Produce wrong, not just slow, downstream behavior

## Requested behavior

1. **Never emit empty stdout on a successful API call.** Even for a genuinely empty result, return a structurally valid JSON envelope (`{"items": []}`, `{"labels": []}`, etc.) that downstream consumers can parse.
2. **Exit non-zero when output serialization fails.** If the formatter cannot produce output for a response it received, that is an error condition — exit with a distinct non-zero code and print a descriptive message to stderr including the endpoint and relevant debug information.
3. **Consider a `--raw` or `--debug-response` flag** that bypasses the formatter and prints the raw API response, so users can recover data even when the formatter fails on a specific payload shape.

## Proposed fix direction

Audit the output-formatting pipeline for silently-caught errors. In Rust, this often looks like `.ok()` or `let _ = ...` on a `Result` where `?` or `.unwrap_or_else(|e| log_and_exit(e))` would be correct. A grep for `.ok()` on serialization or formatting results would be a reasonable starting point.

Integration tests that would catch this class of bug:

- `labels list` against an account with at least one user label (assert: exit 0 AND stdout is non-empty AND parses as JSON AND `.labels | length >= 1`)
- `messages get` against a message with a complex multipart structure
- `searchContacts` with a query matching at least one known contact

hoyt-harness:
## Problem

Closes #740 (partial — format-collision sub-issue)

`gmail +read` defined a local `--format` arg with `default_value("text")` and
`value_parser(["text", "json"])`. The root `--format` flag is declared
`.global(true)`, so clap propagated the `"text"` default up to the root
`ArgMatches`. `main.rs` reads the global `--format` value, gets `"text"`,
doesn't recognise it as a valid output format (valid: `json`, `table`, `yaml`,
`csv`), and emits a spurious warning on every `+read` invocation:

```
warning: unknown output format 'text'; falling back to json (valid options: json, table, yaml, csv)
```

## Fix

Rename the `+read` local arg from `--format` to `--body-format`. The name
collision is eliminated; the global `--format` flag is now unambiguous when
`+read` is the active subcommand.

Behaviour is otherwise unchanged:

| Flag | Effect |
|---|---|
| *(omit)* | Render decoded message body as plain text (default) |
| `--body-format text` | Same as above |
| `--body-format json` | Return the full structured JSON object |

The `after_help` example is updated from `--format json` to `--body-format json`.

## Migration note

Scripts or aliases using `gws gmail +read --format json` must be updated to
`gws gmail +read --body-format json`.

## Testing

- `cargo clippy -- -D warnings`: clean
- `cargo test`: 699 pass, 2 fail (pre-existing `test_get_quota_project_*` ADC
  environment tests unrelated to this change — confirmed failing on unmodified
  `upstream/main` too)

Also collapses a pre-existing `clippy::collapsible_match` in `helpers/script.rs`
that surfaces on every fresh branch from `upstream/main` (same fix as PRs #744,
#747, #748, #749, #750).

### DIFF
--- a/.changeset/fix-read-format-collision.md
+++ b/.changeset/fix-read-format-collision.md
@@ -1,0 +1,19 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Fix `+read` local `--format` arg colliding with the global `--format` flag.
+
+Previously, `gmail +read` defined a local `--format` arg (values: `text`, `json`,
+default: `text`). Because the global `--format` flag is declared with
+`.global(true)`, clap propagated the `text` default up to the parent matches,
+causing `main.rs` to emit a spurious `"unknown output format 'text'; falling back
+to json"` warning on every `+read` invocation.
+
+The local arg is renamed to `--body-format` to eliminate the name collision.
+`--body-format text` (default) renders the decoded message body as plain text;
+`--body-format json` returns the full structured JSON object, matching the previous
+`--format json` behaviour.
+
+**Migration note:** Scripts or aliases using `gws gmail +read --format json` must
+be updated to `gws gmail +read --body-format json`.
--- a/crates/google-workspace-cli/src/helpers/gmail/mod.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/mod.rs
@@ -1808,9 +1808,9 @@ Use fragment tags (<p>, <b>, <a>, etc.) — no <html>/<body> wrapper needed.
                         .action(ArgAction::SetTrue),
                 )
                 .arg(
-                    Arg::new("format")
-                        .long("format")
-                        .help("Output format (text, json)")
+                    Arg::new("body-format")
+                        .long("body-format")
+                        .help("Body rendering format: text (default) or json")
                         .value_parser(["text", "json"])
                         .default_value("text"),
                 )
@@ -1831,7 +1831,7 @@ Use fragment tags (<p>, <b>, <a>, etc.) — no <html>/<body> wrapper needed.
 EXAMPLES:
   gws gmail +read --id 18f1a2b3c4d
   gws gmail +read --id 18f1a2b3c4d --headers
-  gws gmail +read --id 18f1a2b3c4d --format json | jq '.body'
+  gws gmail +read --id 18f1a2b3c4d --body-format json | jq '.body_text'
 
 TIPS:
   Converts HTML-only messages to plain text automatically.
--- a/crates/google-workspace-cli/src/helpers/gmail/read.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/read.rs
@@ -35,7 +35,7 @@ pub(super) async fn handle_read(
         fetch_message_metadata(&client, &t, message_id).await?
     };
 
-    let format = matches.get_one::<String>("format").unwrap();
+    let format = matches.get_one::<String>("body-format").unwrap();
     let show_headers = matches.get_flag("headers");
     let use_html = matches.get_flag("html");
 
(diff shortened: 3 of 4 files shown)


hoyt-harness:
Fixed — the example now uses `.body_text` which is the correct field name on the serialized struct. Good catch.

hoyt-harness:

PR #758, opened 2026-04-25
fix: replace nonexistent Forms API list method with Drive files list

xiaolai:

The `recipe-collect-form-responses` skill references `gws forms forms list` in step 1. The Google Forms API v1 does not have a `list` method on the `forms` resource — it only supports `get`, `create`, `patch`, `batchUpdate`, and `responses.list`. Calling this command will always return an API error, making it impossible for users to discover their form IDs through the recipe.

The fix replaces step 1 with a Drive API files list query filtered to `application/vnd.google-apps.form` MIME type, which correctly enumerates all Google Forms in the user's Drive. Since this now depends on the Drive API, `gws-drive` is added as a prerequisite skill in both the YAML frontmatter `skills:` list and the `PREREQUISITE` line in the body.

### DIFF
--- a/.changeset/fix-collect-form-responses-list.md
+++ b/.changeset/fix-collect-form-responses-list.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Fix recipe-collect-form-responses to use Drive API to list forms (Forms API has no list method)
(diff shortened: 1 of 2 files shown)


xiaolai:
@googlebot I signed it!

xiaolai:
Addressed gemini-code-assist's review: switched the Drive API `q` literal to single quotes per Drive API spec, and added `trashed = false` to filter deleted forms. Force-pushed; the diff is a single-line change to `skills/recipe-collect-form-responses/SKILL.md` line 25.

PR #759, opened 2026-04-25
fix: add --locked flag to cargo install cargo-llvm-cov in coverage.sh

xiaolai:

`scripts/coverage.sh` installs `cargo-llvm-cov` via `cargo install cargo-llvm-cov` without the `--locked` flag. Without `--locked`, Cargo resolves the newest compatible versions of all transitive dependencies at install time rather than using the versions pinned in the tool's own `Cargo.lock`. This means:

1. Different runs of the coverage script can produce different tool binaries depending on what dependency versions are available at that moment.
2. Newly released (and potentially vulnerable or breaking) patch/minor versions of transitive dependencies can be silently pulled in.

Adding `--locked` ensures the install is fully reproducible and uses only the dependency graph the `cargo-llvm-cov` maintainers tested against.

### DIFF
--- a/.changeset/fix-pin-cargo-llvm-cov.md
+++ b/.changeset/fix-pin-cargo-llvm-cov.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Add --locked flag to cargo install cargo-llvm-cov in coverage.sh for reproducible installs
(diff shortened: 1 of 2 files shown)


xiaolai:
@googlebot I signed it!

xiaolai:
Addressed gemini-code-assist's review: pinned `cargo-llvm-cov` to `--version 0.8.5` (current stable on crates.io). Force-pushed; the diff is a single-line change to `scripts/coverage.sh` line 21.

PR #760, opened 2026-04-25
fix: pin devDependencies to exact versions in package.json

xiaolai:

`package.json` uses caret ranges (`^`) for both `@changesets/cli` and `lefthook` in `devDependencies`. Caret ranges allow any compatible minor or patch version to be installed without an explicit update commit. This means:

1. A `pnpm install` in a fresh environment (e.g., a new developer machine or a CI ephemeral runner without a cache) can silently pull in a newer version than the one tested.
2. Supply-chain updates — including potentially malicious patch releases — could reach developer environments and CI without a code review.

Pinning to exact versions (`2.29.8` and `2.1.2`) ensures that every dependency upgrade is a deliberate, reviewable change. The existing `pnpm-lock.yaml` already records exact resolved versions; this change makes `package.json` consistent with that intent.

### DIFF
--- a/.changeset/fix-pin-dev-deps.md
+++ b/.changeset/fix-pin-dev-deps.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Pin devDependencies to exact versions to prevent unreviewed minor/patch updates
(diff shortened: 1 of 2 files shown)


xiaolai:
@googlebot I signed it!

PR #784, opened 2026-05-10
fix(mcp): map 127.0.0.1 bind to localhost in server_base

shigechika:


shigechika:
Wrong target repository — this fix is specific to the shigechika/gws-mcp fork.

PR #841, opened 2026-06-10
fix(gmail): match message headers case-insensitively (#642)

### LINKED ISSUE #642: bug(gmail): parse_message_headers uses case-sensitive match, drops CC/headers with non-canonical casing

lucaspretti:
## Summary

`gws gmail +reply-all` silently drops `Cc` recipients when the original message's `Cc` header arrives with non-canonical casing (e.g. `CC`), which is common with Microsoft Exchange / Outlook. The original `To` recipients survive but the entire `Cc` list is lost, so a reply-all can go out missing people from the thread with no error.

## Root cause

`parse_message_headers` (`crates/google-workspace-cli/src/helpers/gmail/mod.rs`) matched header names with an exact, case-sensitive `match`. Per RFC 5322 §1.2.2 header field names are case-insensitive, and the Gmail API preserves the original casing from the sending MTA, so `CC` (and any other non-canonical casing) fell through to `_ => {}` and was silently dropped.

## Fix

Match on the lowercased header name so all casings are recognized. This also folds the previous `"Message-ID" | "Message-Id"` special-case into a single `"message-id"` arm.

## Test

Adds `test_parse_original_message_case_insensitive_headers`, which feeds Exchange-style casing (`FROM`, `TO`, `CC`, `cc`, `message-id`) and asserts every recipient and the message id are parsed (i.e. `Cc` is not dropped).

Fixes #642

### DIFF
--- a/.changeset/gmail-header-case-insensitive.md
+++ b/.changeset/gmail-header-case-insensitive.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Fix Gmail message header parsing to be case-insensitive (RFC 5322 §1.2.2). `+reply-all` no longer silently drops `Cc` recipients (or other headers) from Microsoft Exchange / Outlook senders that use non-canonical casing such as `CC`.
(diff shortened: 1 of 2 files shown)


lucaspretti:
@googlebot I signed it!

lucaspretti:
@googlebot I signed it!

lucaspretti:
@googlebot I signed it!

PR #847, opened 2026-06-18
fix(gmail): match message header names case-insensitively (#642)

### LINKED ISSUE #642: bug(gmail): parse_message_headers uses case-sensitive match, drops CC/headers with non-canonical casing

shigechika:
## Summary

Fixes #642.

`parse_message_headers` matched header field names with exact case:

```rust
match name {
    "From" => ...
    "Cc" => ...
    ...
}
```

Gmail preserves the sender's original header casing, so a message whose `Cc`
header arrives as `"CC"` (very common from Exchange/Outlook) — or `"from"`
lowercase from some MTAs — falls through the match and is **silently dropped**.
In practice this drops all CC recipients from `gws gmail +reply-all`.

Per [RFC 5322 §1.2.2](https://datatracker.ietf.org/doc/html/rfc5322#section-1.2.2) header field names are case-insensitive. The sibling `get_part_header` already uses `eq_ignore_ascii_case`, so this just brings `parse_message_headers` in line.

## Fix

Normalize the name to lowercase before matching:

```rust
    "from" => ...
    "cc" => ...
    ...
}
```

(The previous `"Message-ID" | "Message-Id"` arm collapses to a single
`"message-id"`.)

## Test

Adds `test_parse_message_headers_case_insensitive`, covering mixed-case
`FROM` / `to` / `CC` / `Reply-TO` / `subject` / `MESSAGE-ID` / `References`.

`cargo build` clean; `cargo test helpers::gmail` green (235 passed).

### DIFF
--- a/.changeset/fix-gmail-header-case-insensitive.md
+++ b/.changeset/fix-gmail-header-case-insensitive.md
@@ -1,0 +1,12 @@
+---
+"@googleworkspace/cli": patch
+---
+
+fix(gmail): match message header names case-insensitively.
+
+`parse_message_headers` used exact-case string matching, so headers whose field
+names use non-canonical casing — e.g. `"CC"` (common from Exchange/Outlook) or a
+lowercase `"from"` from some MTAs — fell through and were silently dropped. This
+dropped CC recipients from `+reply-all`. Per RFC 5322 §1.2.2 header field names
+are case-insensitive (the sibling `get_part_header` already uses
+`eq_ignore_ascii_case`). Normalize the name to lowercase before matching.
--- a/crates/google-workspace-cli/src/helpers/gmail/mod.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/mod.rs
@@ -258,15 +258,20 @@ fn parse_message_headers(headers: &[Value]) -> ParsedMessageHeaders {
         let name = header.get("name").and_then(|v| v.as_str()).unwrap_or("");
         let value = header.get("value").and_then(|v| v.as_str()).unwrap_or("");
 
-        match name {
-            "From" => parsed.from = value.to_string(),
-            "Reply-To" => append_address_list_header_value(&mut parsed.reply_to, value),
-            "To" => append_address_list_header_value(&mut parsed.to, value),
-            "Cc" => append_address_list_header_value(&mut parsed.cc, value),
-            "Subject" => parsed.subject = value.to_string(),
-            "Date" => parsed.date = value.to_string(),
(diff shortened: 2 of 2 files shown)


lucaspretti:
Hey @shigechika, thanks for picking this up. I submitted the same fix in #846 (a re-submission of #841), but the stale bot closed both before a maintainer could take a look.

Since the diff here is essentially identical, two small things from my version might be worth folding in:

- a regression test (`test_parse_original_message_case_insensitive_headers`) that feeds Exchange-style casing (`FROM`, `TO`, `CC`, `cc`, `message-id`) and asserts no recipient or the message id gets dropped;
- the old `"Message-ID" | "Message-Id"` arm collapsed into a single `"message-id"` case.

No ownership concerns on my end, just flagging so whichever PR lands keeps the test coverage. Glad to see this getting fixed either way.

atze1210 (approved):

shigechika:
Thanks @lucaspretti — sorry the stale bot got both of yours. Good news: both points you raised are already covered here:

- the old `"Message-ID" | "Message-Id"` arm is collapsed into a single `"message-id"` case;
- `test_parse_message_headers_case_insensitive` feeds mixed Exchange-style casing (`FROM`/`to`/`CC`/`Reply-TO`/`MESSAGE-ID`/`References`) and asserts none of them are dropped.

Genuinely happy for whichever PR lands — the important thing is that #642 gets fixed. Thanks for flagging the test coverage either way.

shigechika:
Still relevant — #642 (dropped CC recipients from case-sensitive header matching) isn't fixed on main yet. Reopening to keep this available for review; happy to keep it open until a maintainer can take a look.

shigechika:
Couldn't get this one to reopen via the API/UI (GitHub kept rejecting it), so opened a fresh PR with the identical diff instead: #869. No code changes — just a re-submission so a maintainer can still review it.

PR #865, opened 2026-07-05
fix: preserve unreserved path ID characters

### LINKED ISSUE #842: `script scripts run` percent-encodes unreserved characters in scriptId path parameter, causing 404
When calling `gws script scripts run`, the `scriptId` passed via `--params` is percent-encoded in the URL path, turning `-` into `%2D` and `_` into `%5F`. These are unreserved characters per RFC 3986 and should not be encoded. The resulting URL is rejected by the Apps Script Execution API with 404.

## Steps to reproduce

```bash
gws script scripts run \
  --params '{"scriptId":"1-ukYIb8XjT9KQlsj6_NbryM5UJXBwAMeY5GKtSBA05csXty3iY7DhO-P"}' \
  --json '{"function":"myFunction","devMode":true}' \
  --dry-run
```

## Expected URL

```
https://script.googleapis.com/v1/scripts/1-ukYIb8XjT9KQlsj6_NbryM5UJXBwAMeY5GKtSBA05csXty3iY7DhO-P:run
```

## Actual URL

```
https://script.googleapis.com/v1/scripts/1%2DukYIb8XjT9KQlsj6%5FNbryM5UJXBwAMeY5GKtSBA05csXty3iY7DhO%2DP:run
```

**Result:** API returns `404 Requested entity was not found`, even though the same `scriptId` works correctly with `gws script projects get`.

Note: `gws script projects get --params '{"scriptId":"..."}' ` works fine — the path encoding issue appears to be specific to the `scripts run` endpoint.

## Version

gws 0.22.5

zerone0x:
## Summary
Fixes #842

Preserve RFC 3986 unreserved characters in path parameter encoding so Apps Script IDs containing `-` and `_` stay valid in `script scripts run` URLs.

## Changes
- Replace the overly strict `NON_ALPHANUMERIC` encode set with a path-segment encode set that leaves `-`, `.`, `_`, and `~` intact.
- Keep URL-structure characters such as `/`, `?`, `#`, `%`, `@`, and `=` percent-encoded.
- Add a regression test using the scriptId shape from the issue.

## Test Plan
- `cargo fmt -- --check`
- `cargo test -q validate::tests::test_encode_path_segment_preserves_unreserved_script_id` (blocked in this environment: linker `cc` is not installed, so Rust build scripts for dependencies fail before tests compile)
- `cargo test -q` (same linker blocker)


zerone0x:
Superseded by #867, which is rebased onto upstream main to avoid the fork branch conflict.

PR #873, opened 2026-07-12
docs: document Gmail web link vs API ID mismatch (#858)

### LINKED ISSUE #858: Docs: no supported way to build a Gmail web link from an API ID (inverse of #790)
## Use case

This is the inverse of #790. Given an email found through the API, I want to build a clickable Gmail web URL that opens that exact message in the browser, so I can cite it in notes, summaries, or replies. This matters most for agents, which reference an email by the IDs the API hands back.

## The gap

The API returns only hex IDs (`id`, `threadId`). Both ways to turn one into a URL fall short:

1. `https://mail.google.com/mail/u/0/#all/<hexId>` opens the message only in a Gmail tab that is already loaded and signed in, where the web app resolves the hex as a legacy alias. On a cold click it fails.
2. `#search/rfc822msgid:<Message-ID>` works, but it is a search URL, not a direct link to the message.

A durable link needs Gmail's opaque web ID (`FMfcgz…`, `Ktbx…`, `CXKn…`, `QgrcJ…`). The API never returns it, and #790 shows it cannot be derived offline because it is wrapped with a per-account server key. So the only reliable path today is to open each message in a browser and copy the address bar, which defeats automation.

## What I verified

I decoded four permalinks from one account. The hex `id` and `threadId` appear nowhere in the bytes. Messages in the same thread share a leading block (the thread part) and differ in the tail (the message part), and that thread block is longer than the 8-byte hex ID, which fits an encrypted token. This matches the decode in #790.

## Request

1. Document, in the README and the `gws-gmail` skill, that no supported method turns an API hex ID into a durable web link. The `#all/<hexId>` alias is unofficial and breaks on a cold load; an `rfc822msgid` search is the only API-derivable link. This is the forward-direction counterpart to #790, which covers web URL to API ID.
2. If Google later exposes a permalink, return it on message and thread reads.
3. Cross-link #790.

## Environment

- gws 0.22.5
- Node v24.16.0
- Used through the agent skills (`gws-gmail`)

MeGaurav4:
Reopening. CLA has been signed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ gws drive --help      # shows +upload …
 | `modelarmor` | `+sanitize-response` | Sanitize a model response through a Model Armor template |
 | `modelarmor` | `+create-template` | Create a new Model Armor template |
 
-> **Gmail web links vs API IDs:** The API returns hex thread/message IDs (e.g. `19e1acd465710c93`). There is no supported way to turn these into durable Gmail web URLs. `#all/<hexId>` is unofficial and unreliable on cold loads. For the inverse problem (web URL IDs not accepted by the API), see [#790](https://github.com/googleworkspace/cli/issues/790).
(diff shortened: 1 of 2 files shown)


MeGaurav4:
reopen

MeGaurav4:
magically reopen

PR #885, opened 2026-07-28
feat(auth): customize the OAuth callback for remote-host logins

npeshkov:
## Description

`gws auth login`'s flow assumes the browser can reach the CLI's callback server on `localhost`.

When the CLI runs on a remote host (e.g. a cloud development environment), the server is not reachable from the developer's laptop browser, so the OAuth callback has to be redirected to an address other than the laptop's `localhost`.

This Pull Request adds three optional environment variables that let the caller customize the OAuth flow:

- `GOOGLE_WORKSPACE_CLI_OAUTH_REDIRECT_URI` - sets the redirect URI sent to Google.
- `GOOGLE_WORKSPACE_CLI_OAUTH_STATE` - a `state` value passed to the authorization URL (see [use cases](https://auth0.com/docs/secure/attack-protection/state-parameters)).
- `GOOGLE_WORKSPACE_CLI_OAUTH_PORT` - sets the port of the local callback listener.

## Implementation notes

- The authorisation code could be percent-encoded on the way between Google's server and the remote host. This will lead to `invalid_grant: Malformed auth code`, unless the code is decoded by `gws`.
- When `GOOGLE_WORKSPACE_CLI_OAUTH_STATE` is set, the `state` returned on the callback is verified against it. A missing or mismatched value is rejected to guard against forged (CSRF) callbacks.
- The callback server bounds how much it reads from the request line (8 KiB), so a client that never sends a newline cannot exhaust memory (DoS) - relevant now that the port may be reachable from remote networks.
- Moved helper for testing with environment variables from `auth.rs` to `test_support.rs`.

## Checklist

### DIFF
--- a/.changeset/oauth-redirect-override.md
+++ b/.changeset/oauth-redirect-override.md
@@ -2,4 +2,4 @@
 "@googleworkspace/cli": minor
 ---
 
-Allow customizing the `gws auth login` OAuth callback so it works when the CLI runs on a remote host whose `localhost` the browser connot reach directly: `GOOGLE_WORKSPACE_CLI_OAUTH_REDIRECT_URI` overrides the redirect URI sent to Google, `GOOGLE_WORKSPACE_CLI_OAUTH_STATE` sets the `state` value passed through to the auth URL, and `GOOGLE_WORKSPACE_CLI_OAUTH_PORT` pins the local callback (loopback) port.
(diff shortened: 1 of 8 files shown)


npeshkov:
Hey @jpoehnelt not sure if you are still active in `gws` but this tool is great and I'd love to make it even better with this tiny PR. Could you help me find someone with the time to take a look at it? 🙏

PR #893, opened 2026-08-05
feat: add read-only shared hook state verifier

mandalorianuy:
## Summary
- add a read-only `gws shared-hook-state` verifier that resolves the shared Git common directory and fails closed on missing, invalid, or drifting authority
- add a versioned human decision contract plus governance documentation for forward hook state without claiming historical restoration
- add unit and CLI coverage for valid, undecided, missing, malformed, symlink, permission, and drift cases

## Validation
- `cargo test --quiet` — 798 passed, 0 failed
- `cargo fmt --all -- --check`
- `cargo clippy -- -D warnings`
- `cargo build`
- `git diff --check origin/main...HEAD`
- `./target/debug/gws shared-hook-state` — PASS, DECIDED, VALID, drift none, mutation none

## Delivery note
The local pre-push wrapper printed `Cannot find lefthook in PATH` and did not block the push. This is not counted as hook validation; the full Rust suite and the independent read-only verifier above are the technical evidence.

### DIFF
--- a/.changeset/shared-hook-state-authority.md
+++ b/.changeset/shared-hook-state-authority.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Add a read-only, fail-closed authority contract and verifier for shared Git hook state.
--- a/.changeset/shared-hook-state-authority.md
+++ b/.changeset/shared-hook-state-authority.md
@@ -2,4 +2,4 @@
 "@googleworkspace/cli": patch
 ---
 
-Add a read-only, fail-closed authority contract and verifier for shared Git hook state.
+Add a read-only, fail-closed authority contract, exact forward decision, and verifier for shared Git hook state.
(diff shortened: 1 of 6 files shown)


mandalorianuy:
Fail-closed post-push update: the local pre-push bootstrap regenerated the shared `pre-commit`, `pre-push`, and `lefthook.checksum` files after the successful pre-push validation. A fresh read-only verification now returns `status=DRIFT`, `errorCode=TARGET_DRIFT`, `exitCode=23`, with hash mismatches on all three targets. The wrappers now point at the isolated T9 worktree dependency path rather than the previously approved checkout path. No hook restoration or contract update was attempted because the current human decision governs exact bytes. This PR is being kept as draft pending a new explicit forward-state decision. Separate remote gates also remain: Google CLA is unsigned and external-contributor workflows require maintainer approval.

mandalorianuy:
Closing this proposal because it modeled clone-local Codex/Lefthook governance as Google Workspace CLI product work. The incident was introduced by Codex while managing shared hooks across linked worktrees; it is not a gws product defect and does not belong in the upstream command surface. Containment and verification are being handled exclusively in clone-local Git administrative state. No product remediation is requested from maintainers.

PR #897, opened 2026-08-12
fix(gmail): preserve rich message bodies in read

### LINKED ISSUE #889: bug(gmail): `+read` silently loses body content — prefers empty text/plain part, truncates, drops anchor hrefs
`gws gmail +read` returns incomplete message bodies with no indication that content was dropped. Three distinct failures, all silent:

**1. Prefers a stub `text/plain` part over a rich `text/html` part.** On a real `multipart/alternative` message, the text part was 59 chars with zero URLs; the HTML part was 17,590 chars containing 6 links. `+read` returned the 59-char stub, discarding the entire message.

**2. Truncates.** A 3,209-byte draft came back as 198 bytes, cut mid-word. No `[truncated]` marker.

**3. Drops anchor targets.** HTML→text conversion discards `href` values, so any message whose payload *is* a link — invitations, password resets, portal access, OTP magic links — is unreadable. One invitation rendered as invisible word-joiner characters with no URL at all.

Also: HTML entities aren't unescaped (`&#39;` appears literally) and block elements don't produce line breaks, so multi-paragraph bodies collapse to one line.

**Impact:** consumers can't tell content was lost. Worse for drafts — a correct outbound draft *looks* mangled, and the natural response is to delete and rebuild it.

**Suggested:** emit the full decoded body, select the richer alternative part when text/plain is empty or substantially shorter, preserve href targets (inline or as a link list), unescape entities, and if any cap is retained, print an explicit `[truncated N bytes]` marker.

Version: 0.22.5

jokrasno:
## Description

Fixes #889.

Gmail `+read` currently falls back to the truncated snippet for HTML-only messages and can prefer a short or empty `text/plain` alternative over substantially richer HTML. Naive tag stripping also loses link targets, entity decoding, and block structure.

This change:

- renders HTML bodies to readable text with `html2text`, retaining links and document structure
- prefers rendered HTML only when it is substantially richer than a non-empty plain-text alternative
- keeps comparable valid plain text and uses the Gmail snippet only when neither body yields readable content
- adds a patch changeset and locks the new dependency

All fixtures are synthetic; no Gmail account, credentials, or real messages were used.

## Validation

The five initial regressions all failed on pristine `a3768d0`:

```text
0 passed; 5 failed; 701 filtered out
```

They cover richer multipart selection, non-truncation, anchor `href` preservation, HTML entity decoding, and block line breaks. The final suite also covers empty and comparable plain-text alternatives.

- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p google-workspace-cli test_read_` — 7 passed
- [x] `cargo test -p google-workspace-cli helpers::gmail::` — 242 passed
- [x] `cargo build --workspace`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] serialized workspace suite — CLI 707 passed with two unrelated, pre-existing Windows ADC tests excluded
- [x] Changeset included

The two excluded ADC tests mutate `HOME` while `dirs::home_dir()` reads `USERPROFILE` on Windows; they are outside the Gmail helper and unchanged by this commit.

### DIFF
--- a/.changeset/fix-gmail-read-body-extraction.md
+++ b/.changeset/fix-gmail-read-body-extraction.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Preserve complete, structured Gmail message content when `+read` converts HTML bodies to text.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -929,6 +929,7 @@ dependencies = [
  "futures-util",
  "google-workspace",
  "hostname",
+ "html2text",
  "iana-time-zone",
  "keyring",
  "mail-builder",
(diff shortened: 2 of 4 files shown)


jokrasno:
The Google CLA is now signed and the CLA check passes. The branch is still current with main and GitHub reports it mergeable. The normal author reopen operation is returning a validation error after the stale bot closed it; could a maintainer please reopen this PR? I’m ready to address review or CI feedback.

PR #923, opened 2026-09-09
fix(gmail): preserve Cc recipients regardless of header casing

### LINKED ISSUE #911: gmail +reply-all drops the original message's CC recipients (v0.22.5)
**Version:** 0.22.5 (Homebrew bottle, macOS arm64)

**Repro:** reply to a message whose headers are `From: A`, `To: me`, `CC: B, C` with:

```
gws gmail +reply-all --message-id <id> --body "..."
```

**Expected** (reply-all semantics): outgoing message addressed to A with B and C kept on CC (all original recipients minus self).

**Actual:** the sent message has `To: A` only — the CC header is dropped entirely, silently. Verified by fetching the sent message's metadata headers immediately after. The helper accepts `--cc` for *additional* recipients, but the whole point of reply-all is that the original CC set carries without the caller re-supplying it.

Happy to provide sanitised header dumps if useful.

ZIFeIYUuuuuuu:
Fix Gmail +reply-all dropping Cc recipients when header casing differs

## Problem

Gmail header names are case-insensitive, but `+reply-all` only recognized the exact spelling `Cc`. Messages returned with `CC`, `cc`, or another casing were parsed without any Cc recipients, so the reply omitted people who were copied on the original message.

Minimal reproduction:

1. Reply-all to a message whose Gmail API payload contains a header such as `{ "name": "CC", "value": "alice@example.com" }`.
2. The current parser falls through to `_` and leaves `ParsedMessageHeaders.cc` empty.
3. The generated reply therefore does not include the original Cc recipient.

## Root cause

`parse_message_headers` used a case-sensitive string match for `"Cc"`. RFC 5322 header field names are case-insensitive, and Gmail can return equivalent casing variants.

## Changes

- Match the Cc header with `eq_ignore_ascii_case("Cc")` while preserving existing address-list concatenation and formatting.
- Add a focused regression test covering both `CC` and `cc` headers and asserting that both recipients are retained.
- Add the required patch changeset.

## Validation

- `cargo test -p google-workspace-cli helpers::gmail::tests -- --nocapture` (113 passed)
- `cargo test -p google-workspace-cli helpers::gmail::tests::test_parse_message_headers_preserves_cc_regardless_of_header_case -- --exact --nocapture` (1 passed)
- `cargo fmt --all -- --check` (passed)
- `cargo clippy -p google-workspace-cli -- -D warnings -A clippy::collapsible-match` (passed; the allow is for an existing unrelated warning in `helpers/script.rs`)
- `git diff --check` (passed)

Other header handling and normal truncation/error behavior are unchanged. This closes #911.

### DIFF
--- a/.changeset/fix-gmail-cc-header-case.md
+++ b/.changeset/fix-gmail-cc-header-case.md
@@ -1,0 +1,5 @@
+---
+"@googleworkspace/cli": patch
+---
+
+Preserve Gmail Cc recipients when the API returns a differently cased header name.
--- a/crates/google-workspace-cli/src/helpers/gmail/mod.rs
+++ b/crates/google-workspace-cli/src/helpers/gmail/mod.rs
@@ -262,7 +262,9 @@ fn parse_message_headers(headers: &[Value]) -> ParsedMessageHeaders {
             "From" => parsed.from = value.to_string(),
             "Reply-To" => append_address_list_header_value(&mut parsed.reply_to, value),
             "To" => append_address_list_header_value(&mut parsed.to, value),
-            "Cc" => append_address_list_header_value(&mut parsed.cc, value),
+            _ if name.eq_ignore_ascii_case("Cc") => {
+                append_address_list_header_value(&mut parsed.cc, value)
+            }
             "Subject" => parsed.subject = value.to_string(),
             "Date" => parsed.date = value.to_string(),
             "Message-ID" | "Message-Id" => parsed.message_id = value.to_string(),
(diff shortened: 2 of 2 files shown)


ZIFeIYUuuuuuu:
This was auto-closed by the inactivity bot, and the CI runs at that time were all still at `action_required` rather than having executed. Their job count is 0 and `updated_at` is just the close timestamp, so the `failure` conclusions are an artefact of the PR being closed, not real test results.

What actually happened before close:

- `cla/google` passed ✅
- `googleworkspace-bot` ran `/gemini review` on 2026-09-09
- but the PR never got a human look, and `CI` / `Coverage` / `Release` / `Policy` / `Actions Workflow Security Scan` never started

The change itself is small (`+20/-1`, 2 files): thread the original header casing through when preserving Cc recipients instead of matching on a normalised key.

Happy to keep it current with `main` while it's open. Could someone approve the workflow runs so CI can actually run?

PR #927, opened 2026-09-10
fix: honor explicit version in `<api>:<version>` syntax for unlisted services

JamCodex685:
## Summary

The CLI's own error message advertises `<api>:<version>` as an escape hatch for services not in the hardcoded `SERVICES` table (e.g. `chromepolicy:v1`), but that syntax never actually worked: `parse_service_and_version` always calls `resolve_service()` unconditionally, which has no fallback and only matches the hardcoded list. The version parsed out of the colon syntax was silently discarded whenever the service name wasn't already a known alias.

The Discovery-document fetch underneath (`discovery.rs`) is already generic over arbitrary `(service, version)` pairs, so the fix is contained to the CLI arg-resolution layer: when `resolve_service` fails and the colon syntax was used, fall back to treating the literal strings as a Discovery API name/version pair instead of erroring.

## Test plan

- Added 3 unit tests: a colon-qualified unlisted service now resolves, a known alias still resolves via the existing path (no regression), and an unlisted service *without* a colon still errors (no accidental loosening of validation).
- `cargo test`: 712 passed, 0 failed.
- `cargo clippy --all-targets -- -D warnings`: clean for the changed code (11 pre-existing errors on `main` in unrelated files — not introduced by this change).
- `cargo fmt --check`: clean.
- Live smoke test against two real, previously-unreachable APIs:
  - `gws 'customsearch:v1' --help` — builds the full `cse` resource command tree instead of "Unknown service".
  - `gws 'chromepolicy:v1' customers policySchemas --help` — builds the real `get`/`list` command tree.

Changeset included.


JamCodex685:
I signed it!

